PR #1, opened 2015-03-13
Adding initial v2 work

jamiehannaford:
The basic structure for this codebase is vaguely similar to v1, but with a few differences:

1. API data, such as which parameters an operation accepts, what the HTTP verb is, etc. is placed in `Api` classes for each service. Each operation has a static method inside this class, and returns an array. 
2. All REST logic is handled by an `Operation` class. The service methods (`createServer`) will instantiate an `Operation`, pass in the array from `OpenStack\Compute\v2\Api::postServers`, and the Operation will handle the HTTP transaction, not the service. This decouples the service layer from REST and HTTP stuff, which I quite like. It will also avoid unnecessary duplication.
3. Authentication is handled by attaching a listener to the Guzzle HTTP client. Every time a HTTP request is sent, a `before` event is fired off - and our auth handler listens out for it. When this happens, it checks the current stored token to see if its valid, and if not generates a new one. It then sets the `X-Auth-Token` header for the HTTP request.
4. Resources are split into two things: `OperatorResource` and `ValueResource`. An `Operator` is anything that can execute REST operations; a `ValueResource` is just a representation of data. A service is also an `Operator`.
5. We use interfaces a lot more, and traits
6. I chose to use phpspec instead of phpunit - and this is a controversial one. I've given up with behat because it's way too much overhead and a bit orthogonal to our needs for this OSS project. phpspec, however, has worked very nicely because it allows us way more control modelling objects and controlling how they interact with other objects. For integration tests, I think we should rely on sample scripts.

## Questions
1. Are we happy with phpspec or shall we use phpunit?
2. Are the `Api` classes a good solution for representing API data?
3. Are we happy to use public properties for model data. I used to hate the idea, but the more I think about it, there's no real difference to user between `$server->name` and `$server['name']`. In fact, public properties can be seen by IDEs.


samples/compute/v2/create_server.php
```diff
@@ -0,0 +1,16 @@
...
+use OpenStack\OpenStack;
+
+$openstack = new OpenStack;
+
+$compute = $openstack->computeV2([
+    'region' => 'RegionOne',
+    'debug'  => true,
+]);
+
+$s = $compute->createServer('php_test', 'e37365c2-5c45-4b73-b4ae-828436d5c569', '1');
```
ycombinator: Should we take in multiple parameters or one parameter that is an associative array for future-proofing?

ycombinator:
I've started looking at this but it's going to take me some time just to get a lay of the land - how the classes are laid out, how they interact with each other, etc. I'll probably start off with a few questions about why things are the way they are so I can understand your motivations before making any comments.

src/Compute/v2/Api/Server.php
```diff
@@ -0,0 +1,204 @@
...
+                'limit'        => ['type' => 'string', 'location' => 'query'],
+                'status'       => ['type' => 'string', 'location' => 'query'],
+                'host'         => ['type' => 'string', 'location' => 'query']
+            ],
+        ];
+    }
+
+    public static function getAllDetail()
+    {
+        $definition = self::getServers();
```
ycombinator: Where is `getServers()` defined?

src/Compute/v2/Api/Server.php
```diff
@@ -0,0 +1,204 @@
...
+                        ]
+                    ],
+                ],
+            ]
+        ];
+    }
+
+    public static function getAll()
+    {
+        return [
```
ycombinator: This structure reminds me a lot of WADL or any of the more modern API description vocabularies. Could we somehow use those instead of coming up with our own?

src/Compute/v2/Api/Server.php
```diff
@@ -0,0 +1,204 @@
...
+        return $definition;
+    }
+
+    public static function get()
+    {
+        return [
+            'method' => 'GET',
+            'path'   => 'servers/{serverId}',
+            'params' => [
+                'id' => ['type' => 'string', 'required' => true, 'location' => 'url']
```
ycombinator: Should this `id` really be `serverId` (or vice versa)? Otherwise I'm confused how/where they are being mapped to each other.

src/Compute/v2/Models/Server.php
```diff
@@ -0,0 +1,71 @@
...
+        $response = $this->execute(ServerApi::put(), $this->getAttrs(['id', 'ipv4', 'ipv6']));
+
+        $this->fromResponse($response);
+    }
+
+    public function delete()
+    {
+        $this->execute(ServerApi::delete(), $this->getAttrs(['id']));
+    }
+}
```
ycombinator: I notice that this `Server` class implements `create()`, `update()`, and `delete()` as thin wrappers around `ServerApi::post()`, `ServerApi::put()`, and `ServerApi::delete()` respectively. How do the other methods in `ServerApi` (namely, `changePassword()` through `createImage()`) get exposed to the user of the SDK?

src/Compute/v2/Service.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace OpenStack\Compute\v2;
+
+use OpenStack\Common\Service\AbstractService;
+
+class Service extends AbstractService
+{
+    public function createServer($name, $imageId, $flavorId, array $options = [])
```
ycombinator: Are you following an idiom of have required parameters be positional and the last positional parameters be an array for all optional parameters?

src/Common/Resource/ResourceInterface.php
```diff
@@ -0,0 +1,35 @@
...
+     * @param array $data
+     * @return mixed
+     */
+    public function populateFromArray(array $data);
+
+    /**
+     * @param ResponseInterface $response
+     * @return mixed
+     */
+    public function setLastResponse(ResponseInterface $response);
```
ycombinator: Is the purpose of this method to allow debugging the response, if necessary?

src/Compute/v2/Models/Server.php
```diff
@@ -0,0 +1,71 @@
...
+
+use OpenStack\Compute\v2\Api\Server as ServerApi;
+use OpenStack\Common\Resource\AbstractResource;
+use OpenStack\Common\Resource\IsCreatableInterface;
+use OpenStack\Common\Resource\IsDeletableInterface;
+use OpenStack\Common\Resource\IsUpdateableInterface;
+use OpenStack\Common\Resource\OperatorResource;
+use OpenStack\Common\Resource\ResourceInterface;
+
+class Server extends OperatorResource implements ResourceInterface,
```
ycombinator: As `OperatorResource` already implements `ResourceInterface`, does `Server` need to implement it too?

src/Compute/v2/Api/Server.php
```diff
@@ -0,0 +1,204 @@
...
+            ]
+        ];
+    }
+
+    public static function getAll()
+    {
+        return [
+            'method' => 'GET',
+            'path'   => 'servers',
+            'params' => [
```
ycombinator: This structure reminds me a lot of JSON Schema. Could we somehow re-use that instead of coming up with our own vocabulary?

samples/compute/v2/create_server.php
jamiehannaford: I was 100% for assoc arrays before, but actually they're not great for IDE users - positional args can be a lot more explicit. Then again, we can mitigate this issue by documenting the parameters well.

Do you want to use assoc arrays for _all_ methods? It seems a bit verbose to ask users to do:

```php
$s = $compute->getServer(['id' => 'e37365c2-5c45-4b73-b4ae-828436d5c569']);
```

Maybe we could have a general rule of thumb: positional arguments can be used for <= 3 required parameters. For >3 required parameters, and for optional parameters, use an associative array. What do you think?

src/Compute/v2/Api/Server.php
jamiehannaford: Whoops, this should be `self::getAll` :)

src/Compute/v2/Api/Server.php
jamiehannaford: Yeah, totally. It's not 100% similar, but it uses some of the structure and definitions sparingly. When you say "use existing vocabularies", do you mean that we should replace our use of terminology with the WADL equivalents, effectively keeping our current implementation? Or do you think we should overhaul our implementation to actually use WADL/JSON schema?

The problem I have with WADL and any other vocabularies is that they get very verbose and offer us a bunch of functionality we will never need in the SDK. Also some of things we actually need have no equivalent in JSON schema or WADL (like aliasing for example). I want a more focused vocabulary that is lean and precise. An operation only has four keys:

`path` - the URL path
`method` - HTTP method
`jsonKey` - the top-level JSON key for request body (i.e. `{"server": {}}`)
`params` - an array of user parameters

Each parameter is nested, and has these keys:

`type` - the value type (string, integer, boolean)
`required` - whether or not the user must provide a value
`location` - either `url`, `json`, `body`, `header` or `query`
`sentAs` - the alias. So `contentType` will be sent as `Content-Type`
`path` - used for JSON bodies. it will insert the value into the JSON structure using an XPath-like definition.

Also, if we used WADL or JSON schemas, we'd probably need an external parsing library - which just adds to our dependency list - which I definitely don't want. I only think we need the above definitions, so IMHO it's not a major issue - we're just formalizing the stuff we'd want to store anyway.

src/Compute/v2/Api/Server.php
jamiehannaford: yeah, it should be - mistake on my part

src/Compute/v2/Models/Server.php
jamiehannaford: we'd just need to add those methods - I haven't added them yet

src/Common/Resource/ResourceInterface.php
jamiehannaford: yeah, that's right. I think it would be useful if people could access the last response from a model. what do you think?

src/Compute/v2/Models/Server.php
jamiehannaford: nope, good catch

jamiehannaford:
Thanks for the great feedback @ycombinator. I'm going to merge this and implement all the stuff we talked about directly.

PR #2, opened 2015-05-20
Fix errors in CONTRIBUTING

notFloran:
Fix table of "Running integration tests"
Replace "rackspace/gophercloud" with "php-opencloud/openstack"


jamiehannaford:
Great catch, thanks! :rocket:

PR #4, opened 2015-05-25
Add the keypair attribute for createServer

notFloran:


jamiehannaford:
Thanks for taking a look at this. The reason I didn't add keypairs to the v2 API class is that it's technically considered an extension (`os-keypairs`). For extensions, I want to keep them slightly separate from the base API classes, since we shouldn't force them on users who haven't adopted them - the base API class should be the common denominator.

So far, I haven't spent much time thinking how we can structure extensions. We can probably place the API definitions in a separate namespace like `Compute\v2\Exts\Keypairs`, but I don't know how we can decorate the base models (like `Compute\v2\Models\Server`) to include any added properties. Do you have any ideas?

notFloran:
Okay, good idea to separate the extensions of the base API.

I will think about a solution with the help of @nclsHart

PR #5, opened 2015-06-19
Add "Networking" service

notFloran:
Manage *Network* of the *Networking* service.

Todo : 
- [x] Create network
- [x] Bulk create networks
- [x] Show network
- [x] Update network
- [x] Delete network

### DIFF
--- a/src/Networking/v2/Models/Network.php
+++ b/src/Networking/v2/Models/Network.php
@@ -39,6 +39,17 @@ class Network extends AbstractResource implements Listable, Retrievable
     /**
      * {@inheritDoc}
      *
+     * @param array $userOptions {@see \OpenStack\Networking\v2\Api::postNetworks}
+     */
+    public function bulkCreate(array $userOptions)
+    {
+        $response = $this->execute($this->api->postNetworks(), $userOptions);
+        return $this->populateFromResponse($response);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
      * @param array $userOptions {@see \OpenStack\Networking\v2\Api::postNetwork}
      */
     public function create(array $userOptions)
--- a/src/Networking/v2/Models/Network.php
+++ b/src/Networking/v2/Models/Network.php
@@ -43,7 +43,9 @@ class Network extends AbstractResource implements Listable, Retrievable
      */
     public function bulkCreate(array $userOptions)
     {
-        $response = $this->execute($this->api->postNetworks(), $userOptions);
+        $response = $this->execute($this->api->postNetworks(), [
+            'networks' => $userOptions,
+        ]);
         $body = $response->json();
         $json = $body['networks'];
 
--- a/src/Networking/v2/Models/Network.php
+++ b/src/Networking/v2/Models/Network.php
@@ -4,6 +4,8 @@ namespace OpenStack\Networking\v2\Models;
 
 use OpenStack\Common\Resource\AbstractResource;
 use OpenStack\Common\Resource\Listable;
+use OpenStack\Common\Resource\Creatable;
+use OpenStack\Common\Resource\Deletable;
 use OpenStack\Common\Resource\Retrievable;
 
 /**
@@ -11,7 +13,7 @@ use OpenStack\Common\Resource\Retrievable;
  *
  * @property \OpenStack\Networking\v2\Api $api
  */
-class Network extends AbstractResource implements Listable, Retrievable
+class Network extends AbstractResource implements Listable, Retrievable, Creatable, Deletable
 {
     public $id;
     public $name;
@@ -39,12 +41,12 @@ class Network extends AbstractResource implements Listable, Retrievable
     /**
      * {@inheritDoc}
      *
-     * @param array $userOptions {@see \OpenStack\Networking\v2\Api::postNetworks}
+     * @param array $data {@see \OpenStack\Networking\v2\Api::postNetworks}
      */
-    public function bulkCreate(array $userOptions)
+    public function bulkCreate(array $data)
     {
         $response = $this->execute($this->api->postNetworks(), [
-            'networks' => $userOptions,
+            'networks' => $data,
         ]);
         $body = $response->json();
         $json = $body['networks'];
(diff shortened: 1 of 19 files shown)


jamiehannaford:
@notFloran :+1: Give me a shout when you want this reviewed

notFloran:
@jamiehannaford Ok !

I have a problem, since the last commits I can't authenticate, I get this error : 
```
Uncaught exception 'InvalidArgumentException' with message 'Either a user or token must be provided.'
```

It seems that with Identity v3, *OS_USER_ID* and *OS_PROJECT_ID* are needed but where I can get them ?

jamiehannaford:
@notFloran I recently added [sample files](/php-opencloud/openstack/samples/identity/v3) for v3 identity. You'll need to do something like this:

```php
$openstack = new OpenStack\OpenStack([
    'authUrl' => '{authUrl}',
    'region'  => '{region}',
    'user'    => [
        'id'       => '{userId}',
        'password' => '{password}'
    ],
    'scope' => [
        'project' => [
            'id' => '{projectId}'
        ]
    ]
]);
```

You'll need to retrieve the `{projectId}` value from your OpenStack installation. The easiest way to do this is to log in to the Horizon control panel as an admin, and find the project your API user will be using.

notFloran:
Ok, thanks I found UserID and ProjectID.

I also have to change the auth URL to use Identity v3 (https://XXX/v3), now I get authenticated but all url are prefixed with /v3.

Example : 
```
POST /v3/v2.0/networks HTTP/1.1
...
HTTP/1.1 404 Not Found
```

If I let my default auth Url (https://XXX/v2.0), the Identity v3 service use the wrong url to authenticate. Example : 
```
POST /v2.0/auth/tokens HTTP/1.1
...
HTTP/1.1 404 Not Found
```

jamiehannaford:
@notFloran After authenticating with keystone v3, does your catalog have the correct service URLs? All the identity service does is extract from the catalog, so I don't know why it's prepending `v3` :confused:

notFloran:
@jamiehannaford I made a PR to fix that : #6

notFloran:
For the *Bulk create networks*, I don't know how handle this type of request : 
```
{
    "networks": [
        {
            "name": "sample_network3",
            "admin_state_up": true
        },
        {
            "name": "sample_network4",
            "admin_state_up": true
        }
    ]
}
```

For the moment, I'm stuck with this:
```
        return [
            'path' => $this->pathPrefix . '/networks',
            'method' => 'POST',
            'jsonKey' => 'networks',
            'params' => [
                'network' => [
                    'type' => 'array',
                    'description' => 'List of networks',
                    'items' => [
                        'type'       => 'object',
                        'properties' => [
                            'name' => $this->nameParam,
                            'shared' => $this->sharedParam,
                            'adminStateUp' => $this->adminStateUp,
                        ]
                    ],
                ]
            ]
        ];
```

```
{"networks":{"network":[{"name":"networkName1"},{"name":"networkName2"}]}}
```

How to give an array of objects to the **params** of *Operation* ?

jamiehannaford:
@notFloran Will that definition not work? This is how I'd call it:

```php
$networks = [
   ['name' => 'foo', 'shared' => true],
   ['name' => 'bar', 'shared' => false],
];

$service->bulkCreate($networks);
```

notFloran:
@jamiehannaford I think it's good for review

src/Networking/v2/Models/Network.php
```diff
@@ -0,0 +1,90 @@
...
+use OpenStack\Common\Resource\AbstractResource;
+use OpenStack\Common\Resource\Listable;
+use OpenStack\Common\Resource\Retrievable;
+
+/**
+ * Represents a Network v2 Network.
+ *
+ * @property \OpenStack\Networking\v2\Api $api
+ */
+class Network extends AbstractResource implements Listable, Retrievable
```
jamiehannaford: this should also implement `Creatable` and `Updateable` I think

src/Networking/v2/Models/Network.php
```diff
@@ -0,0 +1,90 @@
...
+
+        return $networks;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param array $userOptions {@see \OpenStack\Networking\v2\Api::postNetwork}
+     */
+    public function create(array $userOptions)
```
jamiehannaford: when you implement the interface, you will need to rename this arg

src/Networking/v2/Models/Network.php
```diff
@@ -0,0 +1,90 @@
...
+        $response = $this->execute($this->api->postNetwork(), $userOptions);
+        return $this->populateFromResponse($response);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function update()
+    {
+        $response = $this->execute($this->api->putNetwork(), $this->getAttrs(['id', 'name', 'shared', 'adminStateUp']));
```
jamiehannaford: I added a new method called [`executeWithState`](https://github.com/php-opencloud/openstack/blob/master/src/Common/Resource/AbstractResource.php#L172-L176) that will look at what params the API definition wants, and then populate an array of those keys based off the current object's state. It basically means we can delegate to this method and not have to manually define the update keys. So you'd call like this:

```php
$response = $this->executeWithState($this->api->putNetwork());
return $this->populateFromResponse($response);
```

src/Networking/v2/Service.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function createNetwork(array $options)
+    {
+        return $this->model('Network')->create($options);
+    }
+
+    /**
+     * Create a new network resources.
+     *
+     * @param array $options {@see \OpenStack\Networking\v2\Api::postsNetwork}
```
jamiehannaford: I think this should be `:: postNetworks`

src/Networking/v2/Service.php
```diff
@@ -0,0 +1,54 @@
...
+     * Retrieve a network object without calling the remote API. Any values provided in the array will populate the
+     * empty object, allowing you greater control without the expense of network transactions. To call the remote API
+     * and have the response populate the object, call {@see Server::retrieve}.
+     *
+     * @param array $options An array of attributes that will be set on the {@see Network} object. The array keys need to
+     *                       correspond to the class public properties.
+     *
+     * @return \OpenStack\Networking\v2\Models\Network
+     */
+    public function getNetwork(array $options = [])
```
jamiehannaford: how do you feel about having a string `$networkId` as the arg, rather than an associative array? I know it's not that consistent, but I think it'll be easier for users to access instead of constructing `['id' => 'foo']` all the time. what do you think?

jamiehannaford:
@notFloran Overall this is great, I just have a few very minor issues. once they're addressed I'll happily merge :shipit:

notFloran:
@jamiehannaford I fix this minor issues.

For the *getNetwork* : I think getNetwork($id) is better for the user.

jamiehannaford:
Thanks again @notFloran :rocket:

PR #6, opened 2015-06-23
Fix auth with Identity v3

notFloran:
If Identity service is created in setupHttpClient of the builder, set catalogName and catalogType to false as is done in [OpenStack](https://github.com/php-opencloud/openstack/blob/master/src/OpenStack.php#L72).

The method getUrl of Service, check name only is name is set in the Service. In the OpenStack doc, **name** is marked as optional.


jamiehannaford:
:+1: good catch

PR #8, opened 2015-07-28
Add Subnet to networking

notFloran:


notFloran:
@jamiehannaford I think it's good for review

src/Networking/v2/Models/Subnet.php
```diff
@@ -0,0 +1,104 @@
...
+     * {@inheritDoc}
+     */
+    public function retrieve()
+    {
+        $response = $this->execute($this->api->getSubnet(), ['id' => (string) $this->id]);
+        $this->populateFromResponse($response);
+    }
+
+    /**
+     * {@inheritDoc}
```
jamiehannaford: we don't have a `bulkCreate` in `AbstractResource`, so we should define a brief description of what this operation does and its return type (`[]Subnet`)

src/Networking/v2/Models/Subnet.php
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @param array $data {@see \OpenStack\Networking\v2\Api::postSubnets}
+     */
+    public function bulkCreate(array $data)
+    {
+        $response = $this->execute($this->api->postSubnets(), [
+            'subnets' => $data,
+        ]);
+        $body = $response->json();
```
jamiehannaford: how about:

```php
$response = $this->execute($this->api->postSubnets(), ['subnets' => $data]);
$subnets = $response->json()['subnets'];
```

src/Networking/v2/Models/Subnet.php
```diff
@@ -0,0 +1,104 @@
...
+        $response = $this->executeWithState($this->api->putSubnet());
+        return $this->populateFromResponse($response);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function delete()
+    {
+        $this->execute($this->api->deleteSubnet(), $this->getAttrs(['id']));
```
jamiehannaford: nothing wrong with this, but I always use `executeWithState` whenever possible, since all necessary attributes will be picked up according to the API definition, rather than us manually remembering. Not really a big issue here since it's obvious that `id` is the only attr

src/Networking/v2/Params.php
```diff
@@ -33,4 +33,113 @@ public function adminStateUp()
...
+            'description' => 'The CIDR',
+        ];
+    }
+
+    public function tenantId()
+    {
+        return [
+            'type'        => self::STRING_TYPE,
+            'sentAs'      => 'tenant_id',
+            'description' => 'The ID of the tenant who owns the network. Only administrative users can specify a tenant ID other than their own. You cannot change this value through authorization policies',
```
jamiehannaford: :+1: we should not be afraid of writing long descriptions, since it will always help the user when they're reading the reference docs. for anything longer than the standard page width, use [heredoc syntax](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc)

src/Networking/v2/Service.php
```diff
@@ -48,4 +48,42 @@ public function getNetwork($id)
...
+    {
+        return $this->model('Subnet')->create($options);
+    }
+
+    /**
+     * Create a new subnet resources.
+     *
+     * @param array $options {@see \OpenStack\Networking\v2\Api::postSubnets}
+     *
+     * @return array
```
jamiehannaford: `[]Subnet`

src/Networking/v2/Service.php
```diff
@@ -48,4 +48,42 @@ public function getNetwork($id)
     {
         return $this->model('Network', ['id' => $id]);
     }
+
+    /**
+     * Create a new subnet resource.
+     *
+     * @param array $options {@see \OpenStack\Networking\v2\Api::postSubnet}
+     *
+     * @return \OpenStack\Networking\v2\Models\Subnet
```
jamiehannaford: If you're referencing the path namespace as an import path for this class, I don't see any harm in using the shorter version: `Subnet`. The FQCN is absolutely fine though :smile:

tests/integration/Networking/v2Test.php
```diff
@@ -56,6 +71,9 @@ private function createNetworksAndDelete()
         require_once $path;
 
         foreach($networks as $network) {
+            $this->assertInstanceOf('OpenStack\Networking\v2\Models\Network', $network);
```
jamiehannaford: With these types of assertions, one thing I've started to do is:

```php
$this->assertInstanceOf(Network::class, $network);
```

because my IDE will handle writing out the FQCN for me (as an import at the top of the class). Another reason is that if we ever change the namespace path (which has definitely happened in the past), lots of string are harder to update rather than 1 import path at the top of each test file. I found this out in v1!

tests/integration/Networking/v2Test.php
```diff
@@ -126,4 +144,130 @@ private function deleteNetwork()
 
         $this->logStep('Deleted network ID', ['ID' => $this->networkId]);
     }
+
+    private function createSubnetsAndDelete()
+    {
+        $replacements = [
+            '{subnetName1}' => 'fakeSubnet1',
```
jamiehannaford: feel free to use [`$this->randomStr()`](https://github.com/php-opencloud/openstack/blob/master/tests/integration/TestCase.php#L96-L107) helper for string generation. It will also add a common prefix, which is useful when teardowns randomly fail to clean up everything

tests/integration/Networking/v2Test.php
```diff
@@ -126,4 +144,130 @@ private function deleteNetwork()
...
+    private function deleteSubnet()
+    {
+        $replacements = ['{subnetId}' => $this->subnetId];
+
+        /** @var $subnet \OpenStack\Networking\v2\Models\Subnet */
+        $path = $this->sampleFile($replacements, 'delete_subnet.php');
+        require_once $path;
+
+        $this->logStep('Deleted subnet ID', ['ID' => $this->subnetId]);
+    }
```
jamiehannaford: :boom: great job with these tests

tests/unit/Networking/v2/Models/SubnetTest.php
```diff
@@ -0,0 +1,134 @@
...
+                ],
+                [
+                    'name' => $opts[1]['name'],
+                    'network_id' => $opts[1]['networkId'],
+                    'tenant_id' => $opts[1]['tenantId'],
+                    'ip_version' => $opts[1]['ipVersion'],
+                    'cidr' => $opts[1]['cidr'],
+                ],
+            ],
+        ];
```
jamiehannaford: I'm thinking maybe we can create a helper method to generate the expected JSON based off the input options. For example:

```php
$expectedJson = $this->expectedJson($opts, [
   'networkId' => 'network_id',
   'ipVersion'  => 'ip_version'
]);
```

As you can see, aliases are passed in as the second arg. Or, even better, why not pass in the operation data? We'd need to add a new class property:

```php
public function __construct()
{
   // existing stuff
   $this->api = new Api();
   $this->subnet = new Subnet($this->client->reveal(), $this->api);
   $this->subnet->id = '{subnetId}';
}

public function test_it_bulk_creates()
{
   $opts = [];

   $expectedJson = $this->expectedJson($opts, $this->api->postSubnets()['params']);
}
```
I can add this in the future, what do you think? Worth it or not?

samples/networking/v2/create_subnet.php
```diff
@@ -0,0 +1,29 @@
...
+    ]
+]);
+
+$networking = $openstack->networkingV2();
+
+$options = [
+    'name' => '{subnetName}',
+    'networkId' => '{networkId}',
+    'ipVersion' => 4,
+    'cidr' => '192.168.199.0/24'
```
jamiehannaford: this is a real nit pick: but can we align the array keys?

jamiehannaford:
@notFloran This is a great PR, thanks for submitting it. I've posted all my comments as annotations

tests/unit/Networking/v2/Models/SubnetTest.php
notFloran: We definitely need a helper for that :).

I like the idea to use the operation data to create the expected json.

notFloran:
I have fixes all the issues.

jamiehannaford:
Thanks again @notFloran! :shipit:

PR #11, opened 2015-07-29
Fix error in populateFromResponse

notFloran:
**populateFromReponse** now check the content-type for *application/json* but on some responses the content-type is *application/json; charset=UTF-8*


jamiehannaford:
great catch! :+1:

PR #12, opened 2015-07-29
Fix enumerate on several Identity models

notFloran:


jamiehannaford:
Merge all the things :fireworks:

PR #16, opened 2015-08-12
Fix QuerySerializer : use getName instead of $paramName

### LINKED ISSUE #15: Wrong parameter domainId sent to Identity
Hello.

I'm trying to get the projects filtering them by domain_id. As a result i receive all the projects (no filtering) from Keystone.
In logs i see this:
2015-08-12 02:30:39.035 17756 INFO keystone.common.wsgi [-] GET /projects?**domainId**=a40e2b97749a41e58e23ad60e421559b

So the problem is that Keystone receives "domainId" and not "domain_id".

Here is the source code:
```php
<?php
require 'openstack/vendor/autoload.php';

$openstack = new OpenStack\OpenStack([
        'authUrl' => 'http://test.local:35357/v3',
        'region' => 'REG1',
        'user' => [
                'name'          => 'admin',
                'password'      => '......',
                'domain'        => [ 'name' => 'default' ]
        ],
        'scope' => [
                'domain' => [
                        'id' => 'default'
                ]
        ]
]);

$client = $openstack->identityV3();

foreach ($client->listProjects(['domainId' => 'a40e2b97749a41e58e23ad60e421559b']) as $project) {
        print $project->name;
        print "\n";
}
```

notFloran:
Fix #15

### DIFF
--- a/src/Common/Api/QuerySerializer.php
+++ b/src/Common/Api/QuerySerializer.php
@@ -32,7 +32,7 @@ class QuerySerializer
                 continue;
             }
 
-            $query->set($paramName, $value);
+            $query->set($schema->getName(), $value);
         }
 
         $url->setQuery($query);


jamiehannaford:
:shipit: Thanks for the fix

PR #18, opened 2015-11-16
Guzzle 6 fix integration tests for v2 Identity

mkissam:
Add Integration test support for the V2 Identity API services, based on the value of OS_IDENTITY_API_VERSION environment variable.


jamiehannaford:
:+1:

PR #20, opened 2015-11-16
Fix class name of specific integration tests

mkissam:
The class name of integration test was not properly converted in case
of object store. The service names were collected from samples directory
where object store was represented as "object_store" meanwhile the
integration test name is called "ObjectStore" in camel-case format.

This fix adds a new camel case formatter that properly converts the
object_store to ObjectStore.

### DIFF
--- a/tests/integration/Runner.php
+++ b/tests/integration/Runner.php
@@ -81,8 +81,9 @@ class Runner
         return $services;
     }
 
-    private function toCamelCase($word, $separator = '_') {
-      return str_replace($separator, '', ucwords($word, $separator));
+    private function toCamelCase($word, $separator = '_')
+    {
+        return str_replace($separator, '', ucwords($word, $separator));
     }
 
     public function runServices()
--- a/tests/integration/Runner.php
+++ b/tests/integration/Runner.php
@@ -81,13 +81,8 @@ class Runner
         return $services;
     }
 
-    private function getCamelCaseStr($className) {
-      $elements = explode('_', $className);
-      $output = '';
-      foreach ($elements as $element) {
-        $output .= ucfirst(trim($element));
-      }
-      return $output;
+    private function toCamelCase($word, $separator = '_') {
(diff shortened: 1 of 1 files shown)


tests/integration/Runner.php
```diff
@@ -81,6 +81,15 @@ private function getRunnableServices($service, $version)
         return $services;
     }
 
+    private function getCamelCaseStr($className) {
```
jamiehannaford: all formatting should be consistent with [PSR-2](http://www.php-fig.org/psr/psr-2/). That means:

- 4 spaces for indents
- opening signature curly brace on separate line for classes and methods

tests/integration/Runner.php
```diff
@@ -81,6 +81,15 @@ private function getRunnableServices($service, $version)
...
+      $elements = explode('_', $className);
+      $output = '';
+      foreach ($elements as $element) {
+        $output .= ucfirst(trim($element));
+      }
+      return $output;
+    }
```
jamiehannaford: how about something like:

```php
private function toCamelCase($word, $separator = '_')
{
    return str_replace($separator, '', ucwords($word, $separator));
}
```

mkissam: Looks much better.

tests/integration/Runner.php
```diff
@@ -81,6 +81,10 @@ private function getRunnableServices($service, $version)
         return $services;
     }
 
+    private function toCamelCase($word, $separator = '_') {
```
jamiehannaford: This does not adhere to [PSR-2 coding style](http://www.php-fig.org/psr/psr-2/). Would you mind taking a quick look at the standard for future reference? :smile: It should be:

```php
private function toCamelCase($word, $separator = '_')
{
    return str_replace($separator, '', ucwords($word, $separator));
}
```

mkissam:
Ah really, forget to run the PSR-2 check...

jamiehannaford:
:rocket:

PR #21, opened 2015-11-16
Refactor normalizeUrl

mkissam:
Refactor private method Builder.normalizeUrl into OpenStack\Common\Transport\Utils as a public static function. This change makes the normalizeUrl callable from external methdos too.


jamiehannaford:
@mkissam :+1: I've merged `guzzle-6` branch into master, so we can point our PRs to that now.

PR #22, opened 2015-11-17
Refactor integration test auth to support v2 identity

mkissam:
Move the integration TestCase auth classes into OpenStack\Integration\Utils. Rewrite the connection substitution code of TestCase.sampleFile to provide v2 auth compatibility for sample classes. Also fix the affected integration tests using the getAuthOpts calls.

### DIFF
--- a/tests/integration/Utils.php
+++ b/tests/integration/Utils.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace OpenStack\Integration;
+
+use GuzzleHttp\Client;
+use OpenStack\Identity\v2\Api;
+use OpenStack\Identity\v2\Service;
+use OpenStack\Common\Transport\HandlerStack;
+
+class Utils
+{
+    public static function getAuthOptsV3()
+    {
+        return [
+            'authUrl' => getenv('OS_AUTH_URL'),
+            'region'  => getenv('OS_REGION_NAME'),
+            'user'    => [
+                'id'       => getenv('OS_USER_ID'),
+                'password' => getenv('OS_PASSWORD'),
+            ],
+            'scope'   => [
+                'project' => [
+                    'id' => getenv('OS_PROJECT_ID'),
+                ]
+            ]
+        ];
+    }
+
+    public static function getAuthOptsV2()
+    {
+        $authUrl = \OpenStack\Common\Transport\Utils::normalizeUrl(getenv('OS_AUTH_URL'));
+        $httpClient = new Client([
+            'base_uri' => $authUrl,
+            'handler'  => HandlerStack::create(),
+        ]);
+        $identityService = new Service($httpClient, new Api);
+        return [
+            'authUrl'         => $authUrl,
+            'region'          => getenv('OS_REGION_NAME'),
+            'username'        => getenv('OS_USERNAME'),
+            'password'        => getenv('OS_PASSWORD'),
+            'tenantName'      => getenv('OS_TENANT_NAME'),
+            'identityService' => $identityService,
+        ];
+    }
+
+    public static function getAuthOpts($options = array())
+    {
+        $auth_options = getenv('OS_IDENTITY_API_VERSION') == '2.0' ?
+            self::getAuthOptsV2() : self::getAuthOptsV3();
+        return array_merge($auth_options, $options);
+    }
+}
--- a/tests/integration/Utils.php
+++ b/tests/integration/Utils.php
@@ -6,6 +6,7 @@ use GuzzleHttp\Client;
 use OpenStack\Identity\v2\Api;
 use OpenStack\Identity\v2\Service;
 use OpenStack\Common\Transport\HandlerStack;
+use OpenStack\Common\Transport\Utils as CommonUtils;
 
 class Utils
 {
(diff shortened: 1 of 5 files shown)


tests/integration/Utils.php
```diff
@@ -0,0 +1,53 @@
...
+                'project' => [
+                    'id' => getenv('OS_PROJECT_ID'),
+                ]
+            ]
+        ];
+    }
+
+    public static function getAuthOptsV2()
+    {
+        $authUrl = \OpenStack\Common\Transport\Utils::normalizeUrl(getenv('OS_AUTH_URL'));
```
jamiehannaford: Can we use `self::normalizeUrl` instead?

tests/integration/Utils.php
```diff
@@ -0,0 +1,53 @@
...
+        $httpClient = new Client([
+            'base_uri' => $authUrl,
+            'handler'  => HandlerStack::create(),
+        ]);
+        $identityService = new Service($httpClient, new Api);
```
jamiehannaford: do we need this as a separate var? can we just put this on line 43?

tests/integration/Utils.php
```diff
@@ -0,0 +1,53 @@
...
+            'username'        => getenv('OS_USERNAME'),
+            'password'        => getenv('OS_PASSWORD'),
+            'tenantName'      => getenv('OS_TENANT_NAME'),
+            'identityService' => $identityService,
+        ];
+    }
+
+    public static function getAuthOpts($options = array())
+    {
+        $auth_options = getenv('OS_IDENTITY_API_VERSION') == '2.0' ?
```
jamiehannaford: `$auth_options` needs to be camelCase :)

tests/integration/Utils.php
```diff
@@ -0,0 +1,53 @@
...
+            'authUrl'         => $authUrl,
+            'region'          => getenv('OS_REGION_NAME'),
+            'username'        => getenv('OS_USERNAME'),
+            'password'        => getenv('OS_PASSWORD'),
+            'tenantName'      => getenv('OS_TENANT_NAME'),
+            'identityService' => $identityService,
+        ];
+    }
+
+    public static function getAuthOpts($options = array())
```
jamiehannaford: `$options = []`

tests/integration/Utils.php
mkissam: I guess it is a different Utils.php not the one under Transports. Would you like to move the auth under Transport Utils or clone the normalizeUrl into this class?

jamiehannaford: You're right, sorry. Can we import the absolute namespace with a `use` statement and just reference the classname on this line then? e.g.

```php
use OpenStack\Common\Transport\Utils;

$authUrl = Utils::normalizeUrl(getenv('OS_AUTH_URL'));
```

mkissam: We cannot, because the class name here is Utils too. I refactor those methods under OpenStack\Common\Transport so it will solve this normalizeUrl call.

jamiehannaford: you need to use an alias:

```php
use OpenStack\Common\Transport\Utils as CommonUtils;
```

tests/integration/Utils.php
```diff
@@ -0,0 +1,53 @@
...
+            'handler'  => HandlerStack::create(),
+        ]);
+        $identityService = new Service($httpClient, new Api);
+        return [
+            'authUrl'         => getenv('OS_AUTH_URL'),
+            'region'          => getenv('OS_REGION_NAME'),
+            'username'        => getenv('OS_USERNAME'),
+            'password'        => getenv('OS_PASSWORD'),
+            'tenantName'      => getenv('OS_TENANT_NAME'),
+            'identityService' => $identityService,
```
jamiehannaford: nitpicking: can we have

```php
'identityService' => new Service($httpClient, new Api),
```

since we're not using `$identityService` anywhere else

tests/integration/Utils.php
```diff
@@ -0,0 +1,53 @@
...
+        ];
+    }
+
+    public static function getAuthOpts($options = [])
```
jamiehannaford: another nitpick: can we have `array $options = []` since it'll only ever be an array

tests/integration/Utils.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $authOptions = getenv('OS_IDENTITY_API_VERSION') == '2.0' ?
+            self::getAuthOptsV2() : self::getAuthOptsV3();
```
jamiehannaford: I'd format it like this:

```php
$authOptions = getenv('OS_IDENTITY_API_VERSION') == '2.0' 
    ? self::getAuthOptsV2() 
    : self::getAuthOptsV3();
```

but I guess it's not important. Also is `2.0` the standard way to define it? Could it ever be `2` or `2.1`?

mkissam: I was following the devstack definition so I guess, 2.0 should be ok, because it is the Identity API Version: http://developer.openstack.org/api-ref-identity-v2.html.

jamiehannaford:
:+1:

PR #25, opened 2015-11-18
add docs for php-opencloud/openstack

alewitt:
added docs for php-opencloud/openstack, previously only the official docs were linked.


jamiehannaford:
Awesome! Thanks for the PR :rocket:

PR #32, opened 2015-11-25
Fix v2 identity private property access bug

mkissam:
The changeset 099ae8b210b3feaa2a1c259f336ac800a6ab9558 broke the v2 identity api, and caused a private property access issue on v2's Entry class. This patchset makes the private properties to
public to make it accessible from AbstractResource.populateFromArray() method.


jamiehannaford:
:shipit:

PR #33, opened 2015-12-29
Added sample in openstack docs for creating an object

karnajani:


jamiehannaford:
@karnajani Thanks! :rocket:

karnajani:
@jamiehannaford what's the difference between this repo and https://github.com/rackspace/php-opencloud?

They are obviously different, just trying to make sure I use the right one for a production app

jamiehannaford:
@karnajani https://github.com/rackspace/php-opencloud is the old v1 library that will be deprecated. This repo is the new redesigned v2, which should be the new official SDK once things are stabilised. This repo has all of the Core OpenStack services, so my feeling is that it can be used in production.

karnajani:
thanks @jamiehannaford 

I appreciate the help :)

PR #38, opened 2016-02-02
Add keyName as an option when creating servers

kunalp:
Add keyName as an option to specify a key pair to be injected into a server upon creation.


jamiehannaford:
@kunalp Thanks for submitting this. Keypairs are _technically_ an OS extension, and I haven't figured out a clean way to implement them yet (that's why we don't have them). But since this is a small fix, I'm happy to merge for now. Thanks again :rocket:

PR #51, opened 2016-04-24
[rfr] Build optimization: Disable xdebug in .travis.yml

haphan:
This will bring build time from 2m58s to 1m37s. @jamiehannaford


jamiehannaford:
👍 Thanks!

PR #61, opened 2016-06-09
[rfr] Map network fields

haphan:
@jamiehannaford

### DIFF
--- a/src/Networking/v2/Models/Network.php
+++ b/src/Networking/v2/Models/Network.php
@@ -39,9 +39,17 @@ class Network extends OperatorResource implements Listable, Retrievable, Creatab
     /** @var string */
     public $tenantId;
 
+    /** @var bool */
+    public $routerExternal;
+
+    /** @var string */
+    public $providerNetworkType;
+
     protected $aliases = [
-        'admin_state_up' => 'adminStateUp',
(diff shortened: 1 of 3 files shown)


jamiehannaford:
Aren't these values provided by the [layer-3](http://developer.openstack.org/api-ref-networking-v2-ext.html#layer3-ext) and [provider](http://developer.openstack.org/api-ref-networking-v2-ext.html#network_provider-ext) extensions?

haphan:
@jamiehannaford these values are retrievable via `/v2.0/networks`, why not also map them to `Network` model? 

Besides, what is the general strategy for sdk to cover extensions?

jamiehannaford:
@haphan All extensions code should live under the `extensions` dir for that service, for example [here](https://github.com/php-opencloud/openstack/tree/master/src/Networking/v2/Extensions). I wanted to do this so that we don't make any assumptions about users using a specific extension and mixing that code with the core service. 

Even if the extension modifies a core URL endpoint (e.g. `/networks`) I still think we should separate but could have the extension resource class (`ExtNetwork`) extend the core resource (`Network`). We could even use the core `Api` and `Params` classes if the operations are identical, but at least we'd have separation.

Issue #64, opened 2016-06-22
Possibility to merge with php-opencloud/common


haphan:
Merging with `php-opencloud/common` will simplify development process and reduce dependencies. 

Can you share what was the reason to separate sdk into 2 repositories @jamiehannaford?

jamiehannaford:
I was thinking the exact same thing. It's a pain to manage 2 repos when 1 would do. Originally I wanted to branch out to allow more providers, but this is now unnecessary. I'm 👍  for merging.

haphan:
Resolved in #65

PR #65, opened 2016-06-22
[rfr] Merge php-opencloud/common

haphan:
https://github.com/php-opencloud/openstack/pull/64


haphan:
@jamiehannaford  shall we also refactor namespace from `OpenCloud` to `OpenStack`?

jamiehannaford:
@haphan Yeah I think that's best if we do

haphan:
@jamiehannaford Any comments?

jamiehannaford:
Awesome work @haphan 🚀

PR #77, opened 2016-08-09
UpApi.php

PandaNiss:
the old file contains no starting or stopping the server

### DIFF
--- a/src/Compute/v2/UpApi
+++ b/src/Compute/v2/UpApi
@@ -1,0 +1,591 @@
+<?php declare (strict_types=1);
+
+namespace OpenStack\Compute\v2;
+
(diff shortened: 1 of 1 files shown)


jamiehannaford:
@PandaNiss Thank you for your contribution. There are some major problems with this PR:

- you do not need to create a new file
- the new file does not have a PHP extension
- there are no tests

Please follow the conventions in the SDK. We have a [contributing guide](https://github.com/php-opencloud/openstack/blob/master/CONTRIBUTING.md) that should help. If you need any questions answered or help, please comment here and I'll reopen this PR. Although I really appreciate the intention, there's not much we can do with this PR so perhaps it's better to start from scratch.

PR #88, opened 2016-08-28
[rfr] Get console vnc/serial/rdp/spice methods

haphan:
@jamiehannaford Please review

### DIFF
--- a/samples/compute/v2/servers/get_server_vnc_console.php
+++ b/samples/compute/v2/servers/get_server_vnc_console.php
@@ -1,0 +1,19 @@
+<?php
+
+require 'vendor/autoload.php';
+
+$openstack = new OpenStack\OpenStack([
+    'authUrl' => '{authUrl}',
+    'region'  => '{region}',
+    'user'    => [
+        'id'       => '{userId}',
+        'password' => '{password}'
+    ],
+    'scope'   => ['project' => ['id' => '{projectId}']]
+]);
+
+$compute = $openstack->computeV2(['region' => '{region}']);
+
+$server = $compute->getServer(['id' => '{serverId}']);
+
+$console = $server->getVncConsole();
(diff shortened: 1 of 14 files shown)


src/Compute/v2/Models/Server.php
```diff
@@ -230,6 +230,59 @@ public function revertResize()
...
 
     /**
+     * Gets a VNC console for a server.
+     *
+     * @param  string $type The type of VNC console: novnc|xvpvnc.
+     *                      Defaults to novnc.
+     *
+     * @return mixed
+     */
+    public function getVncConsole($type = Enum::CONSOLE_NOVNC)
```
jamiehannaford: Why is `$type` an arg here? Surely it'll always be `novnc` for this operation? I like removing all unnecessary expectations for users.

src/Compute/v2/Models/Server.php
```diff
@@ -230,6 +230,59 @@ public function revertResize()
     }
 
     /**
+     * Gets a VNC console for a server.
+     *
+     * @param  string $type The type of VNC console: novnc|xvpvnc.
+     *                      Defaults to novnc.
+     *
+     * @return mixed
```
jamiehannaford: Also, shouldn't the return type be `array` for all these methods?

samples/compute/v2/servers/get_server_vnc_console.php
```diff
@@ -0,0 +1,19 @@
...
+        'password' => '{password}'
+    ],
+    'scope'   => ['project' => ['id' => '{projectId}']]
+]);
+
+$compute = $openstack->computeV2(['region' => '{region}']);
+
+$server = $compute->getServer(['id' => '{serverId}']);
+
+$console = $server->getVncConsole();
```
jamiehannaford: Maybe have `/** @var array $console */`

jamiehannaford:
Looks great, just a few really minor issues

src/Compute/v2/Models/Server.php
haphan: Type is an arg here so that user can pass custom type if needed.

Type can be either `novnc` or `xvpvnc`.  `novnc` seems to be the only type get documented according to [doc](http://developer.openstack.org/api-ref/compute/?expanded=get-vnc-console-os-getvncconsole-action-detail) so I put it as the default parameter.

PR #113, opened 2017-04-28
[rfr] Project quotas

haphan:


haphan:
@jamiehannaford Appreciate if you have time to review this PR.

src/Networking/v2/Models/Quota.php
```diff
@@ -0,0 +1,104 @@
...
+namespace OpenStack\Networking\v2\Models;
+
+use OpenStack\Common\Resource\Deletable;
+use OpenStack\Common\Resource\OperatorResource;
+use OpenStack\Common\Resource\Retrievable;
+use OpenStack\Common\Resource\Updateable;
+use OpenStack\Networking\v2\Api;
+
+/**
+ * Represents a Nova v2 Quota
```
ernesto-pama: Please change Nova to Neutron

src/Compute/v2/Api.php
```diff
@@ -668,4 +668,66 @@ public function getHypervisor(): array
             'params' => ['id' => $this->params->urlId('hypervisor')]
         ];
     }
+
+    public function getQuotaSet()
```
jamiehannaford: needs return types

src/Compute/v2/Api.php
```diff
@@ -668,4 +668,66 @@ public function getHypervisor(): array
...
+                'tenantId' => $this->params->urlId('quota-sets')
+            ]
+        ];
+    }
+
+    public function getQuotaSetDetail()
+    {
+        return [
+            'method'  => 'GET',
+            'path'    => 'os-quota-sets/{tenantId}/detail',
```
jamiehannaford: this is fine, but if you have large operation data structs in the future, it's better to do:

```php
public function getQuotaSetDetail()
{
   $data = self::getQuotaSet();
   $data['url'] += '/detail';
   return $data;
}
```

src/Compute/v2/Models/QuotaSet.php
```diff
@@ -0,0 +1,170 @@
...
+ * Represents a Compute v2 Quota Set
+ *
+ * @property \OpenStack\Compute\v2\Api $api
+ */
+class QuotaSet extends OperatorResource implements Retrievable, Updateable, Deletable
+{
+    /**
+     * The number of allowed instance cores for each tenant.
+     *
+     * @var  int|array
```
jamiehannaford: how can it be both an int and array? then again...openstack 😛

src/Compute/v2/Models/QuotaSet.php
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public $cores;
+
+    /**
+     * The number of allowed fixed IP addresses for each tenant.
+     * Must be equal to or greater than the number of allowed instances.
+     *
+     * @deprecated Since 2.35
+     * @var  int|object
```
jamiehannaford: same here

src/Compute/v2/Models/QuotaSet.php
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public $fixedIps;
+
+    /**
+     * The number of allowed floating IP addresses for each tenant.
+     *
+     * @deprecated Since 2.35
```
jamiehannaford: what does this mean?

src/Compute/v2/Models/QuotaSet.php
```diff
@@ -0,0 +1,170 @@
...
+        'floating_ips'                => 'floatingIps',
+        'injected_file_content_bytes' => 'injectedFileContentBytes',
+        'injected_file_path_bytes'    => 'injectedFilePathBytes',
+        'injected_files'              => 'injectedFiles',
+        'key_pairs'                   => 'keyPairs',
+        'metadata_items'              => 'metadataItems',
+        'security_group_rules'        => 'securityGroupRules',
+        'security_groups'             => 'securityGroups',
+        'server_group_members'        => 'serverGroupMembers',
+        'server_groups'               => 'serverGroups',
```
jamiehannaford: we should really add some kind of conversion logic in Common that handles snake_case -> camelCase

src/Networking/v2/Models/Quota.php
```diff
@@ -0,0 +1,104 @@
...
+     * @var string
+     */
+    public $tenantId;
+
+    /**
+     * @var int
+     */
+    public $subnetpool;
+
+
```
jamiehannaford: no need for extra line

src/Compute/v2/Api.php
haphan: 👍

src/Compute/v2/Api.php
haphan: 👍

src/Compute/v2/Models/QuotaSet.php
haphan: Exactly, but then this is how Openstack implemented it. 

For `os-quota-sets/{tenantId}` returns `int`; `os-quota-sets/{tenantId}/detail` returns an array.  See https://developer.openstack.org/api-ref/compute/?expanded=show-the-detail-of-quota-detail#show-the-detail-of-quota

src/Compute/v2/Models/QuotaSet.php
haphan: See also https://developer.openstack.org/api-ref/compute/?expanded=show-the-detail-of-quota-detail#show-the-detail-of-quota

I think the logic here is that network related quotas such as `floating_ips` and `fixed_ips` shall be managed by Networking service;

I put `@deprecated` flag here to discourage people in using these attributes.

src/Networking/v2/Models/Quota.php
haphan: 👍

src/Compute/v2/Models/QuotaSet.php
jamiehannaford: Hmm, that's a crappy UX for consumers of that API. 

How about we introduce a `Quota` model that encapsulates `in_use`, `limit` and `reserved` fields? If the API returns just an int (e.g. `os-quota-sets/{tenantId}`), that populates just the `limit` field?

src/Compute/v2/Models/QuotaSet.php
jamiehannaford: Ah I see. To clear up confusion, can we expand the comment to:

```
@deprecated Since Nova v2.35. This attribute will eventually move to Neutron, it is advised you do not use this.
```

src/Compute/v2/Models/QuotaSet.php
haphan: We certainly can implement that way, but then at the same time I don't want the SDK to diverse from the API response.

src/Compute/v2/Models/QuotaSet.php
haphan: 👍

src/Compute/v2/Models/QuotaSet.php
haphan: Agree. This issue will be addressed in another PR.

haphan:
@jamiehannaford All good?

jamiehannaford:
@haphan Sure, is your username on there `haphan` too?

PR #114, opened 2017-04-28
[rfr] Allow to update user password

haphan:


haphan:
@jamiehannaford  Please have a look as well.

src/Identity/v3/Api.php
```diff
@@ -540,6 +540,7 @@ public function patchUser(): array
                 'description'      => $this->params->desc('user'),
                 'email'            => $this->params->email(),
                 'enabled'          => $this->params->enabled('user'),
+                'password'         => $this->params->userPassword(),
```
jamiehannaford: Why not use [`$this->params->password()`](https://github.com/haphan/openstack/blob/ec008405be323baba55e99f0e99e01b17b809a42/src/Identity/v3/Params.php#L115-L123)?

PR #124, opened 2017-06-12
[rfr] Set Quota for block storage, fix bug in setting DNS

haphan:


jamiehannaford:
Awesome 👍 Don't have anything to add. Is this ready to merge?

PR #129, opened 2017-07-12
[rtr] Basic Hosts, AZ Support And Other Fixes

clwells:
src/Common/Api/Parameter.php @ 288

This was causing issues noted in commit f377fa58383f8c100495504b890b3e34f0688be0

"This is needed because for the subnet::update() call gatewayIp can be 'null' even though it's a string type.  Null in this case means "remove the gatewayIp"."


haphan:
@clwells Is this ready for review?

clwells:
👍

src/Compute/v2/Models/Server.php
```diff
@@ -314,6 +314,19 @@ public function listAddresses(array $options = []): array
...
     /**
+     * Iterates over all the ports for this server.
+     *
+     * @return array
+     */
+    public function listPorts(array $options = []): array
+    {
+        $options['id'] = $this->id;
+
+        $response = $this->execute($this->api->getPorts(), ['id' => $this->id]);
```
haphan: Instead of `array`, I think it is better to encapsulate in class `InterfaceAttachment`. See [VolumeAttachment](https://github.com/php-opencloud/openstack/blob/master/src/BlockStorage/v2/Models/VolumeAttachment.php) and [Server:: listVolumeAttachments](https://github.com/php-opencloud/openstack/blob/master/src/Compute/v2/Models/Server.php#L434)

src/Compute/v2/Params.php
```diff
@@ -114,7 +114,7 @@ public function imageId(): array
     {
         return [
             'type'        => self::STRING_TYPE,
-            'required'    => true,
+            'required'    => false,
             'sentAs'      => 'imageRef',
             'description' => 'The unique ID of the image that this server will be based on',
```
haphan: Please also update `description` to add clarity when `imageId` field is mandatory.

```The UUID of the image to use for your server instance. This is not required in case of boot from volume. In all other cases it is required and must be a valid UUID```

File: src/Compute/v2/Models/Server.php [hunk 3]
```diff
@@ -314,6 +314,19 @@ public function listAddresses(array $options = []): array
     }
 
     /**
+     * Iterates over all the ports for this server.
+     *
+     * @return array
+     */
+    public function listPorts(array $options = []): array
```
haphan: Use `\Generator`

src/Compute/v2/Api.php
```diff
@@ -649,23 +658,64 @@ public function getHypervisors(): array
...
+                'limit' => $this->params->limit(),
+                'marker' => $this->params->marker()
             ],
         ];
     }
 
     public function getHypervisorsDetail(): array
     {
-        $definition = $this->getHypervisors();
-        $definition['path'] .= '/detail';
```
haphan: Why can't we do this?

src/Common/Api/Parameter.php
```diff
@@ -285,6 +285,11 @@ private function hasCorrectType($userValue): bool
             return true;
         }
 
+        // allow string nulls
+        if ($this->type == 'string' && $userValue === null) {
```
haphan: What is the use case for this? @clwells

clwells: It's mentioned in the commit message for: https://github.com/php-opencloud/openstack/commit/f377fa58383f8c100495504b890b3e34f0688be0

>This is needed because for the subnet::update() call gatewayIp can be 'null' even though it's a string type.  Null in this case means "remove the gatewayIp".

src/Compute/v2/Api.php
clwells: 0 reason apparently :)

clwells:
Everything requested has been done.  LMK if more comes up.

src/Compute/v2/Models/Hypervisor.php
```diff
@@ -62,6 +62,15 @@ class Hypervisor extends OperatorResource implements
     public $vcpus_used;
 
     /** @var array */
+    public $cpu_info;
+
+    /** @var int */
+    public $current_workload;
+
+    /** @var int */
+    public $disk_available_least;
```
jamiehannaford: I'd rather rename everything and release in a minor version bump (v2.1.0). We'd need to document warnings that fields were renamed. Ideally this would be encapsulated in a breaking change release (v3) but I think that's overkill for one class.

@clwells Would you mind updating these fields to camelCame?

.gitignore
```diff
@@ -6,4 +6,5 @@ vendor/
 phpunit.xml
 coverage.xml
 composer.lock
+test
```
jamiehannaford: What is this for?

src/Common/Api/Parameter.php
jamiehannaford: I think this is really inconsistent on OpenStack's part (what's new) but I don't think there's an easier way to do this.

src/Compute/v2/Api.php
```diff
@@ -456,6 +456,18 @@ public function getAddressesByNetwork(): array
         ];
     }
 
+    public function getInterfaceAttachments(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => 'servers/{id}/os-interface',
```
jamiehannaford: minor nit: can we align the array keys

src/Compute/v2/Params.php
```diff
@@ -114,9 +114,9 @@ public function imageId(): array
     {
         return [
             'type'        => self::STRING_TYPE,
-            'required'    => true,
+            'required'    => false,
             'sentAs'      => 'imageRef',
-            'description' => 'The unique ID of the image that this server will be based on',
+            'description' => 'The UUID of the image to use for your server instance. This is not required in case of boot from volume. In all other cases it is required and must be a valid UUID',
```
jamiehannaford: Are we adding extra logic to ensure they specify BFV or an image UUID?

src/Networking/v2/Models/InterfaceAttachment.php
```diff
@@ -0,0 +1,26 @@
...
+    public $id;
+
+    /** @var int */
+    public $device;
+
+    /** @var string */
+    public $serverId;
+
+    /** @var string */
+    public $interfaceId;
```
jamiehannaford: Does this not need aliases? I can't see any tests which verify the HTTP response

src/Compute/v2/Params.php
clwells: What would you like to see here?

clwells:
Beyond my question above all else should be taken care of.

src/Compute/v2/Api.php
```diff
@@ -668,7 +683,40 @@ public function getHypervisor(): array
...
+    }
+
+    public function getAvailabilityZones(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => 'os-availability-zone/detail',
+            'params' => [
+                'limit' => $this->params->limit(),
+                'marker' => $this->params->marker()
```
jamiehannaford: array keys need to be aligned in this file

src/Compute/v2/Params.php
jamiehannaford: Change [this line](https://github.com/nexcess/openstack/blob/master/src/Compute/v2/Api.php#L193) to:

```
$this->notRequired($this->params->imageId())
```

and add logic to [this method](https://github.com/nexcess/openstack/blob/master/src/Compute/v2/Models/Server.php#L107) which ensures either `imageId` or BFV param is provided

src/Networking/v2/Models/InterfaceAttachment.php
```diff
@@ -0,0 +1,49 @@
...
+use OpenStack\Common\Resource\Listable;
+use OpenStack\Common\Resource\OperatorResource;
+
+/**
+ * @property \OpenStack\Networking\v2\Api $api
+ */
+class InterfaceAttachment extends OperatorResource implements Listable
+{
+    /** @var string */
+    public $port_id;
```
jamiehannaford: should be `portId`

tests/unit/Compute/v2/Models/ServerTest.php
```diff
@@ -397,4 +398,13 @@ public function test_it_detaches_volume()
...
         $this->server->detachVolume($attachmentId);
     }
+
+    public function test_it_lists_interface_attachments()
+    {
+        $this->setupMock('GET', 'servers/serverId/os-interface', null, [], 'server-interface-attachments-get');
+
+        $interfaceAttachments = iterator_to_array($this->server->listInterfaceAttachments());
+
+        $this->assertInstanceOf(InterfaceAttachment::class, $interfaceAttachments[0]);
```
jamiehannaford: can we check some field vals too, just to make sure the object is being hydrated properly

clwells:
The squash went.. gross.  See if this is what you wanted.  All changes should have been done but the squash induced some doubt that all got in there tbh.

src/Compute/v2/Params.php
jamiehannaford: L117 needs to stay `true` because it's still used by `rebuildServer`

src/Compute/v2/Models/Host.php
```diff
@@ -0,0 +1,43 @@
...
+use OpenStack\Common\Resource\Retrievable;
+
+/**
+ * Represents a Compute v2 Host.
+ *
+ * @property \OpenStack\Compute\v2\Api $api
+ */
+class Host extends OperatorResource implements Listable, Retrievable
+{
+
```
jamiehannaford: nit: no need for extra return here

src/Compute/v2/Api.php
jamiehannaford: this still needs to be done

src/Compute/v2/Models/Server.php
```diff
@@ -106,6 +107,10 @@ class Server extends OperatorResource implements
      */
     public function create(array $userOptions): Creatable
     {
+        if (!isset($userOptions['imageId']) && !isset($userOptions['blockDeviceMapping']['uuid'])) {
+            throw new \RuntimeException('Boot-from-volume UUID or image UUID is required');
```
jamiehannaford: we need a test for this

jamiehannaford: also can we change the error to `imageId or blockDeviceMapping.uuid must be set`

clwells:
Should be all set.

jamiehannaford:
Thanks for the great work @clwells! 🎉

PR #133, opened 2017-07-19
[rfr] Add swift download options

ekasitk:
This PR allows download options.

# Download the whole object
echo $obj->download()->getContents(); 

# Download only 0-20 bytes with modification time condition
echo $obj->download(['range' => 'bytes=0-20', 'ifModifiedSince' => 'Mon, 17 Jul 2017 07:28:00 ICT'])->getContents();

### DIFF
--- a/src/ObjectStore/v1/Models/Object.php
+++ b/src/ObjectStore/v1/Models/Object.php
@@ -115,11 +115,14 @@ class Object extends OperatorResource implements Creatable, Deletable, HasMetada
 
     public function download(array $data = []): StreamInterface
     {
-        $response = $this->execute($this->api->getObject(), $data + ['containerName' => $this->containerName]);
+        $data += ['name' => $this->name, 'containerName' => $this->containerName];
+
+        /** @var ResponseInterface $response */
+        $response = $this->execute($this->api->getObject(), $data);
         $this->populateHeaders($response);
+
         return $response->getBody();
     }
-
 
     /**
      * {@inheritdoc}
--- a/src/ObjectStore/v1/Models/Object.php
+++ b/src/ObjectStore/v1/Models/Object.php
@@ -112,6 +112,7 @@ class Object extends OperatorResource implements Creatable, Deletable, HasMetada
      *
      * @return StreamInterface
      */
+
     public function download(array $data = []): StreamInterface
     {
         $response = $this->execute($this->api->getObject(), $data + ['containerName' => $this->containerName]);
--- a/src/ObjectStore/v1/Models/Object.php
+++ b/src/ObjectStore/v1/Models/Object.php
@@ -108,15 +108,17 @@ class Object extends OperatorResource implements Creatable, Deletable, HasMetada
      * distinct from fetching its metadata (a `HEAD` request). The body of an object is not fetched by default to
      * improve performance when handling large objects.
      *
+     * @param array $data {@see \OpenStack\ObjectStore\v1\Api::getObject}
+     *
      * @return StreamInterface
      */
-    public function download(): StreamInterface
+    public function download(array $data = []): StreamInterface
     {
-        /** @var ResponseInterface $response */
-        $response = $this->executeWithState($this->api->getObject());
+        $response = $this->execute($this->api->getObject(), $data + ['containerName' => $this->containerName]);
         $this->populateHeaders($response);
         return $response->getBody();
     }
+
 
     /**
      * {@inheritdoc}


src/ObjectStore/v1/Models/Object.php
```diff
@@ -96,11 +96,14 @@ public function retrieve()
...
+     * @param array $options {@see \OpenStack\ObjectStore\v1\Api::getObject} 
+     *
      * @return StreamInterface
      */
-    public function download(): StreamInterface
+    public function download($options=array()): StreamInterface
     {
-        $response = $this->executeWithState($this->api->getObject());
+        $obj = $this->api->getObject();
+        $response = $this->execute($obj, $options + $this->getAttrs(array_keys($obj['params'])));
```
jamiehannaford: Can't this just be

```
$response = $this->execute($this->api->getObject(), $data);
```

src/ObjectStore/v1/Models/Object.php
```diff
@@ -96,11 +96,14 @@ public function retrieve()
      * distinct from fetching its metadata (a `HEAD` request). The body of an object is not fetched by default to
      * improve performance when handling large objects.
      *
+     * @param array $options {@see \OpenStack\ObjectStore\v1\Api::getObject} 
+     *
      * @return StreamInterface
      */
-    public function download(): StreamInterface
+    public function download($options=array()): StreamInterface
```
jamiehannaford: Signature needs to be `array $data = []`

ekasitk: @jamiehannaford   @haphan  Yes, this is fine.

src/ObjectStore/v1/Models/Object.php
ekasitk: @jamiehannaford @haphan 

Not really. We need to use containerName from the previous call to store the container name in the Object.  Actually, all steps look like:

$obj = $swift->getContainer('mycontainer')
                   ->getObject(''test.txt");
echo $obj->download(['range' => 'bytes=0-10'])->getContents();

haphan: @ekasitk I see your point because `containerName` does not exist in `$data`.
To make it consistent

```php
$data['container$options]
$response = $this->execute($this->api->getObject(), $data);
```

src/ObjectStore/v1/Models/Object.php
haphan: Should be `array  $options = []`. I think @jamiehannaford 's points is that we use square bracket for array.

ekasitk:
@haphan 

I don't know how to pass the travis-ci  check. If this is matter, please give me a guide. Thanks.

haphan:
@ekasitk Just pushed a fix for you. Basically in your local, you can execute `php-cs-fixer` like this

```bash
$ php vendor/bin/php-cs-fixer fix --diff --level psr2 -vvv .
```

haphan:
cc @jamiehannaford

src/ObjectStore/v1/Models/Object.php
```diff
@@ -96,11 +96,14 @@ public function retrieve()
      * distinct from fetching its metadata (a `HEAD` request). The body of an object is not fetched by default to
      * improve performance when handling large objects.
      *
+     * @param array $options {@see \OpenStack\ObjectStore\v1\Api::getObject}
+     *
      * @return StreamInterface
      */
-    public function download(): StreamInterface
+    public function download(array $options = []): StreamInterface
```
jamiehannaford: Can we change the variable name to `$data`

src/ObjectStore/v1/Models/Object.php
```diff
@@ -96,11 +96,14 @@ public function retrieve()
...
     {
-        $response = $this->executeWithState($this->api->getObject());
+        $obj = $this->api->getObject();
+        $response = $this->execute($obj, $options + $this->getAttrs(array_keys($obj['params'])));
```
jamiehannaford: Can we change this to `$this->execute($this->api->getObject(), $data);`

haphan: @jamiehannaford Have you seen the justification above?

jamiehannaford: Yeah, sorry @ekasitk I didn't see your comment. What @haphan proposes LGTM because it's consistent with other places in the codebase.

ekasitk: @jamiehannaford  @haphan  OK, I agree with 

$response = $this->execute($this->api-> getObject(), $data + ['containerName' => $this->containerName])

BTW since it's minor change, can @haphan  make this final change? Otherwise, I have to play around with php-cs-fixer for a while.

haphan: @ekasitk Thank you. 
Let's wait for @jamiehannaford  to merge this PR.

PR #134, opened 2017-07-20
[rfr] Fixed stream pointer pointing to end of stream

haphan:
This PR fixes inconsistent behaviour when downloading a file with MIME type of `application/json` using `ObjectStoreV1` service.

`$object->download()` now properly returns Stream with pointer sets at position 0.


haphan:
@jamiehannaford  Please review

Issue #132, opened 2017-07-17
Implement OpenStack's LBAAS


laszlof:
It appears that this library does not support OpenStack's LoadBalancer As A Service functionality. As this is something we plan on using in the near future, I will begin work on implementing this into the current code base.

I wanted to create this issue simply to inform the core team of my intentions to work on this feature. A PR will be submitted once it has been fully tested and is functional.

haphan:
thank you @laszlof

laszlof:
No problem @haphan. Thanks for merging so quickly. Look forward to using this in production.

PR #136, opened 2017-07-21
Add LBAAS API Support

laszlof:
See https://github.com/php-opencloud/openstack/issues/132

This is the initial revision of the LBAAS support for openstack. Theres bound to me somethings in here that need to be changed, but I wanted to get some other eyes on this for the time being.

Please feel free to review/comment/critique and let me know if theres anything I should do differently.

### DIFF
--- a/src/Networking/v2/Api.php
+++ b/src/Networking/v2/Api.php
@@ -350,4 +350,354 @@ class Api extends AbstractApi
             ]
         ];
     }
+
+    public function getLoadBalancers(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/loadbalancers',
+            'params' => []
+        ];
+    }
+
+    public function getLoadBalancer(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/loadbalancers/{id}',
+            'params' => []
+        ];
+    }
+
+    public function postLoadBalancer(): array
+    {
+        return [
+            'method'  => 'POST',
+            'path'    => $this->pathPrefix . '/lbaas/loadbalancers',
+            'jsonKey' => 'loadbalancer',
+            'params'  => [
+                'name'         => $this->params->name('loadbalancer'),
+                'description'  => $this->params->descriptionJson(),
+                'tenantId'     => $this->params->tenantId(),
+                'vipSubnetId'  => $this->params->vipSubnetId(),
+                'vipAddress'   => $this->params->vipAddress(),
+                'adminStateUp' => $this->params->adminStateUp(),
+                'provider'     => $this->params->provider()
+            ]
+        ];
+    }
+
+    public function putLoadBalancer(): array
+    {
+        return [
+            'method'  => 'PUT',
+            'path'    => $this->pathPrefix . '/lbaas/loadbalancers/{id}',
+            'jsonKey' => 'loadbalancer',
+            'params'  => [
+                'id'           => $this->params->idPath(),
+                'name'         => $this->params->name('loadbalancer'),
+                'description'  => $this->params->descriptionJson(),
+                'AdminStateUp' => $this->params->adminStateUp()
+            ]
+        ];
+    }
+
+    public function deleteLoadBalancer() : array
+    {
+        return [
+            'method' => 'DELETE',
+            'path'   => $this->pathPrefix . '/lbaas/loadbalancers/{id}',
+            'params' => [
+                'id' => $this->params->idPath()
+            ]
+        ];
+    }
+
+    public function getLoadBalancerListeners(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/listeners',
+            'params' => []
+        ];
+    }
+
+    public function getLoadBalancerListener(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/listeners/{id}',
+            'params' => []
+        ];
+    }
+
+    public function postLoadBalancerListener(): array
+    {
+        return [
+            'method'  => 'POST',
+            'path'    => $this->pathPrefix . '/lbaas/listeners',
+            'jsonKey' => 'listener',
+            'params'  => [
+                'name'            => $this->params->name('listener'),
+                'description'     => $this->params->descriptionJson(),
+                'loadbalancerId'  => $this->params->loadbalancerId(),
+                'protocol'        => $this->params->protocol(),
+                'protocolPort'    => $this->params->protocolPort(),
+                'tenantId'        => $this->params->tenantId(),
+                'adminStateUp'    => $this->params->adminStateUp(),
+                'connectionLimit' => $this->params->connectionLimit()
+            ]
+        ];
+    }
+
+    public function putLoadBalancerListener(): array
+    {
+        return [
+            'method'  => 'PUT',
+            'path'    => $this->pathPrefix . '/lbaas/listeners/{id}',
+            'jsonKey' => 'listener',
+            'params'  => [
+                'id'              => $this->params->idPath(),
+                'name'            => $this->params->name('listener'),
+                'description'     => $this->params->descriptionJson(),
+                'adminStateUp'    => $this->params->adminStateUp(),
+                'connectionLimit' => $this->params->connectionLimit()
+            ]
+        ];
+    }
+
+    public function deleteLoadBalancerListener() : array
+    {
+        return [
+            'method' => 'DELETE',
+            'path'   => $this->pathPrefix . '/lbaas/listeners/{id}',
+            'params' => [
+                'id' => $this->params->idPath()
+            ]
+        ];
+    }
+
+    public function getLoadBalancerPools(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/pools',
+            'params' => []
+        ];
+    }
+
+    public function getLoadBalancerPool(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/pools/{id}',
+            'params' => []
+        ];
+    }
+
+    public function postLoadBalancerPool(): array
+    {
+        return [
+            'method'  => 'POST',
+            'path'    => $this->pathPrefix . '/lbaas/pools',
+            'jsonKey' => 'pool',
+            'params'  => [
+                'name'               => $this->params->name('pool'),
+                'description'        => $this->params->descriptionJson(),
+                'adminStateUp'       => $this->params->adminStateUp(),
+                'protocol'           => $this->params->protocol(),
+                'lbAlgorithm'        => $this->params->lbAlgorithm(),
+                'listenerId'         => $this->params->listenerId(),
+                'sessionPersistence' => $this->params->sessionPersistence()
+            ]
+        ];
+    }
+
+    public function putLoadBalancerPool(): array
+    {
+        return [
+            'method'  => 'PUT',
+            'path'    => $this->pathPrefix . '/lbaas/pools/{id}',
+            'jsonKey' => 'pool',
+            'params'  => [
+                'id'                 => $this->params->idPath(),
+                'name'               => $this->params->name('pool'),
+                'description'        => $this->params->descriptionJson(),
+                'adminStateUp'       => $this->params->adminStateUp(),
+                'lbAlgorithm'        => $this->params->lbAlgorithm(),
+                'sessionPersistence' => $this->params->sessionPersistence()
+            ]
+        ];
+    }
+
+    public function deleteLoadBalancerPool() : array
+    {
+        return [
+            'method' => 'DELETE',
+            'path'   => $this->pathPrefix . '/lbaas/pools/{id}',
+            'params' => [
+                'id' => $this->params->idPath()
+            ]
+        ];
+    }
+
+    public function getLoadBalancerMembers(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/pools/{poolId}/members',
+            'params' => [
+              'poolId' => $this->params->poolId()
+            ]
+        ];
+    }
+
+    public function getLoadBalancerMember(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/pools/{poolId}/members/{id}',
+            'params' => [
+              'id'      => $this->params->idPath('member'),
+              'poolId' => $this->params->poolId()
+            ]
+        ];
+    }
+
+    public function postLoadBalancerMember(): array
+    {
+        return [
+            'method'  => 'POST',
+            'path'    => $this->pathPrefix . '/lbaas/pools/{poolId}/members',
+            'jsonKey' => 'member',
+            'params'  => [
+                'poolId'       => $this->params->poolId(),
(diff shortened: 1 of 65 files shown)


haphan:
@laszlof Would be nice to cover new code with uni tests.

laszlof:
Yup. Was planning on doing that next. I'm having a bit of trouble with loadbalancer statuses at the moment. The response doesnt seem to want to populate for whatever reason. Debugging now.

Once thats done, I'll go through and add some unit tests.

laszlof:
Have most of the tests done for this. Oddly enough, I found a bug in the HTTP mocking library. If you have a forward slash (`/`) in the value when passing it to the mock, it fails to encode properly which will result in a failed test. I noticed this when setting the `urlPath` option on `LoadbalancerHealthmonitorTest::test_it_creates()`

src/Networking/v2/Api.php
```diff
@@ -350,4 +350,354 @@ public function deleteQuota() : array
...
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/loadbalancers/{loadbalancerId}/statuses',
+            'params' => [
+              'loadbalancerId' => $this->params->loadBalancerIdUrl()
+            ]
+        ];
+    }
+
+    public function getLoadbalancerHealthmonitors(): array
```
haphan: Use camelCase `getLoadBalancerHealthMonitors `

src/Networking/v2/Models/Loadbalancer.php
```diff
@@ -0,0 +1,151 @@
...
+use OpenStack\Common\Resource\Retrievable;
+use OpenStack\Common\Resource\Updateable;
+use OpenStack\Networking\v2\Api;
+
+/**
+ * Represents a Neutron v2 Loadbalancer
+ *
+ * @property Api $api
+ */
+class Loadbalancer extends OperatorResource implements Creatable, Retrievable, Updateable, Deletable
```
haphan: model name should be `LoadBalancer`

src/Networking/v2/Models/Loadbalancer.php
```diff
@@ -0,0 +1,151 @@
...
+     */
+    public $adminStateUp;
+
+    /**
+     * @var string
+     */
+    public $tenantId;
+
+    /**
+     * @var []LoadbalancerListener
```
haphan: put square bracket at the end`LoadBalancerListener[]`. Refer to https://phpdoc.org/docs/latest/guides/types.html#arrays

src/Networking/v2/Models/LoadbalancerHealthmonitor.php
```diff
@@ -0,0 +1,138 @@
...
+use OpenStack\Common\Resource\Retrievable;
+use OpenStack\Common\Resource\Updateable;
+use OpenStack\Networking\v2\Api;
+
+/**
+ * Represents a Neutron v2 Loadbalancer Health Monitor
+ *
+ * @property Api $api
+ */
+class LoadbalancerHealthmonitor extends OperatorResource implements Creatable, Retrievable, Updateable, Deletable
```
haphan: model name `LoadBalancerHealthMonitor `

src/Networking/v2/Models/LoadbalancerHealthmonitor.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    public $adminStateUp;
+
+    /**
+     * @var string
+     */
+    public $poolId;
+
+    /**
+     * @var []LoadbalancerPool
```
haphan: `LoadBalancerPool[]`

src/Networking/v2/Models/LoadbalancerListener.php
```diff
@@ -0,0 +1,137 @@
...
+use OpenStack\Common\Resource\Retrievable;
+use OpenStack\Common\Resource\Updateable;
+use OpenStack\Networking\v2\Api;
+
+/**
+ * Represents a Neutron v2 Loadbalancer Listener
+ *
+ * @property Api $api
+ */
+class LoadbalancerListener extends OperatorResource implements Creatable, Retrievable, Updateable, Deletable
```
haphan: model name `LoadBalancerListener`

src/Networking/v2/Models/LoadbalancerMember.php
```diff
@@ -0,0 +1,117 @@
...
+use OpenStack\Common\Resource\Retrievable;
+use OpenStack\Common\Resource\Updateable;
+use OpenStack\Networking\v2\Api;
+
+/**
+ * Represents a Neutron v2 Loadbalancer member
+ *
+ * @property Api $api
+ */
+class LoadbalancerMember extends OperatorResource implements Creatable, Retrievable, Updateable, Deletable
```
haphan: `LoadBalancerMember`

src/Networking/v2/Models/LoadbalancerPool.php
```diff
@@ -0,0 +1,181 @@
...
+use OpenStack\Common\Resource\Retrievable;
+use OpenStack\Common\Resource\Updateable;
+use OpenStack\Networking\v2\Api;
+
+/**
+ * Represents a Neutron v2 Loadbalancer pool
+ *
+ * @property Api $api
+ */
+class LoadbalancerPool extends OperatorResource implements Creatable, Retrievable, Updateable, Deletable
```
haphan: `LoadBalancerPool`

src/Networking/v2/Models/LoadbalancerPool.php
```diff
@@ -0,0 +1,181 @@
...
+     */
+    public $sessionPersistence;
+
+    /**
+     * @var boolean
+     */
+    public $adminStateUp;
+
+    /**
+     * @var []LoadbalancerListener
```
haphan: `LoadBalancerListener[]`

src/Networking/v2/Models/LoadbalancerPool.php
```diff
@@ -0,0 +1,181 @@
...
+    public function addMember(array $userOptions = []): LoadbalancerMember
+    {
+        $userOptions = array_merge(['poolId' => $this->id], $userOptions);
+        return $this->model(LoadbalancerMember::class)->create($userOptions);
+    }
+
+    /**
+     * Get an instance of a member
+     *
+     * @param string $member_id
```
haphan: use `$id` or `$memberId`

src/Networking/v2/Models/LoadbalancerPool.php
```diff
@@ -0,0 +1,181 @@
...
+     * @return LoadbalancerMember
+     */
+    public function getMember(string $member_id): LoadbalancerMember
```
haphan: use `$id`

src/Networking/v2/Models/LoadbalancerPool.php
```diff
@@ -0,0 +1,181 @@
...
+    {
+        return $this->model(LoadbalancerMember::class, ['poolId' => $this->id, 'id' => $member_id]);
+    }
+
+    /**
+     * Delete a member
+     *
+     * @param string $member_id
+     */
+    public function deleteMember(string $member_id)
```
haphan: use `$id` or `$memberId`

src/Networking/v2/Models/LoadbalancerStat.php
```diff
@@ -0,0 +1,58 @@
...
+
+use OpenStack\Common\Resource\Retrievable;
+use OpenStack\Common\Resource\OperatorResource;
+
+/**
+ * Represents Neutron v2 Loadbalancer Stats
+ *
+ * @property Api $api
+ */
+class LoadbalancerStat extends OperatorResource implements Retrievable
```
haphan: `LoadBalancerStat`

src/Networking/v2/Models/LoadbalancerStat.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    /**
+     * @var integer
+     */
+    public $bytesIn;
+
+    /**
+     * @var integer
```
haphan: I haven't checked API docs but can actual value overflow integer?

src/Networking/v2/Models/LoadbalancerStatus.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public $operatingStatus;
+
+    /**
+     * @var string
+     */
+    public $provisioningStatus;
+
+    /**
+     * @var []LoadbalancerListener
```
haphan: `LoadBalancerListener[]`

src/Networking/v2/Service.php
```diff
@@ -205,4 +211,175 @@ public function getDefaultQuota(string $tenantId): Quota
...
+    {
+        return $this->model(LoadbalancerPool::class)->create($options);
+    }
+
+    /**
+     * Lists loadbalancer members
+     *
+     * @return \Generator
+     */
+    public function listLoadbalancerMembers(string $pool_id): \Generator
```
haphan: `$poolId`

src/Networking/v2/Service.php
```diff
@@ -205,4 +211,175 @@ public function getDefaultQuota(string $tenantId): Quota
...
+
+    /**
+     * Retrieve an instance of a loadbalancer Member object
+     *
+     * @param string $pool_id
+     * @param string $id
+     *
+     * @return LoadbalancerMember
+     */
+    public function getLoadbalancerMember(string $pool_id, string $id): LoadbalancerMember
```
haphan: `$poolId`

haphan:
Following methods should be covered with unit tests as well. 

```
LoadBalancer::addListener
LoadBalancer::getStats
LoadBalancer::getStatuses

LoadbalancerPool::addMember
LoadbalancerPool::getMember
LoadbalancerPool::deleteMember
LoadbalancerPool::addHealthmonitor

LoadbalancerStatus::retrieve
LoadbalancerStat::retrieve
```

src/Networking/v2/Api.php
```diff
@@ -350,4 +350,354 @@ public function deleteQuota() : array
             ]
         ];
     }
+
+    public function getLoadbalancers(): array
```
laszlof: I thought that look better as well, but changed it to match the API docs. In most of the problems I'm working on, you only use capitalization after an `_`. I'll go through and correct this. Thanks.

src/Networking/v2/Models/Loadbalancer.php
laszlof: I copied this format from somewhere else in the project. Will correct.

src/Networking/v2/Models/LoadbalancerStat.php
laszlof: Good point. both bytesIn and bytesOut should probably be float. Do you agree?

laszlof:
Coding style/class name fixes have been taken care of. I'll work on the test coverage either today or tomorrow.

src/Networking/v2/Models/LoadbalancerStat.php
haphan: Openstack uses [BigInt](https://github.com/openstack/neutron-lbaas/blob/9879460e23b84643720337dac7d9c47bedfe0aa1/tools/database-migration-from-v1-to-v2.py#L182) for `bytesIn` and `bytesOut`, unfortunately PHP does not have any equivalent data type.

I personally prefer using `string` in this case since the information here is read-only.

/cc @jamiehannaford

laszlof: We personally implemented our own BigInt class to handle things like this. It was a pain in the butt, but it was a valid workaround at the time. Perhaps we could do something here as well? (or just use one of the `off-the-github` implementations.

I'm not a fan of using a `string` type here, personally.

haphan: @laszlof Coming up with our own type `BigInt` is just too much of reinvent the wheel here. Doctrine also employs `string` for [BigInt](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#bigint).

Secondly, `LoadBalancerStat` only implements `Retrievable` interface which means SDK users cannot mistakenly assign non-numerical character to these fields.

jamiehannaford: Yeah I agree with @haphan. I think `string` is fine for readonly data, because the alternative would involve using a third party dependency for 2 fields. If folks want to do arithmetic on the value, they can use whatever library they want, but I don't think this SDK should enforce that choice on other users.

laszlof: Fair enough. I'll take care of it when I get in the office, and finish the code coverage as well.

laszlof:
@haphan This should be ready to go now.

haphan:
@jamiehannaford  I have no further comments. What do you think?

src/Networking/v2/Models/LoadBalancerStatus.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    private function flattenStatus()
+    {
+        if ($this->loadbalancer instanceof LoadBalancer) {
+            $this->name = $this->loadbalancer->name;
+            $this->id = $this->loadbalancer->id;
+            $this->operatingStatus = $this->loadbalancer->operatingStatus;
+            $this->provisioningStatus = $this->loadbalancer->provisioningStatus;
+            $this->listeners = $this->loadbalancer->listeners;
+            unset($this->loadbalancer);
```
haphan: Why do we need to do this? Is it because `Utils::flattenJson` only accepts resourceKey as a string?

I think having `public $loadbalancer`  then later unset is not a good idea though.

Can we just use do something like

```php
$json = Utils::jsonDecode($response);
$this->populateFromArray(Utils::flattenJson($json, $json[$this->resourceKey]['loadbalancer']));
```
@laszlof

src/Networking/v2/Api.php
```diff
@@ -350,4 +350,354 @@ public function deleteQuota() : array
             ]
         ];
     }
+
+    public function getLoadBalancers(): array
+    {
+        return [
+            'method' => 'GET',
+            'path' => $this->pathPrefix . '/lbaas/loadbalancers',
```
jamiehannaford: nit: we need to make sure all array keys are aligned

src/Networking/v2/Models/LoadBalancerStatus.php
laszlof: Oh, good call. I hadnt thought of that. Will fix.

laszlof:
@jamiehannaford Do you have any docs on creating docs? Want to make sure whatever I write conforms to the standard. Thanks

jamiehannaford:
@laszlof I'd follow the convention in https://github.com/php-opencloud/openstack/tree/master/doc/services. Basically:

1. Create a dir for `load-balancer/<version>`
2. Create a file for each API resource
3. List each operation, and ensure you reference an example file (you will need to create these too)

It's extra work but docs are super important for users. Thanks for your efforts here!

laszlof:
Agreed on docs. I'll try and get to them later today. FWIW, I think the docs should be namespaced under `network`, as it is part of that module.

laszlof:
@jamiehannaford Do you have information on how to build the refdocs? I want to generate them to make sure it looks correct.

jamiehannaford:
@laszlof The refdoc site is currently down, so I wouldn't make it a priority for now.

laszlof:
@jamiehannaford Gotcha. Thanks

laszlof:
Documentation has been added. I chose not to document `LoadBalancerMember` separately, as all of the functionality exists within `LoadBalancerPool` and should be used from this object anyways.

Issue #122, opened 2017-06-12
Unable to get private key when creating keypair


jglatre:
Openstack API reference docs say:
> If you do not provide a public key on create, a new keypair will be built for you, and the private key will be returned during the initial create call.

See: https://developer.openstack.org/api-ref/compute/?expanded=create-flavor-detail,create-remote-console-detail,delete-keypair-detail,create-server-detail,create-or-import-keypair-detail#create-or-import-keypair

But the libray precludes this functionality, AFAIK for two reasons:
1. class `OpenStack\Compute\v2\Models\Keypair` has no `privateKey` property.
2. method `OpenStack\Compute\v2\Params::keypairPublicKey()` declares `publicKey` as required.

zevriev:
I've spend a lot of time to this task on my project.
Try this 

function createKeyPairs() {
        $compute = $this->_openstack->computeV2(['region' => self::$_provider->region]);
        $keys = $this->generateKeys();
        $data = [
            'name'      => $this->_keyName,
            'publicKey' => $keys['public']
        ];
        $keypair = $compute->createKeypair($data);
}        
function generateKeys() {
        $rsa = new Crypt_RSA();
        $rsa->setPublicKey(CRYPT_RSA_PUBLIC_FORMAT_OPENSSH);
        extract($rsa->createKey());
        $rsa->loadKey($publickey);
        $rsa->loadKey($privatekey);
        return array(
            'public' => $rsa->getPublicKey(CRYPT_RSA_PUBLIC_FORMAT_OPENSSH),
            'private'=> $rsa->getPrivateKey(CRYPT_RSA_PUBLIC_FORMAT_OPENSSH)
        );
}

PR #143, opened 2017-08-08
Allow for public keys to be generated by OS, and private key returned

laszlof:
Fixes https://github.com/php-opencloud/openstack/issues/122


jamiehannaford:
Thanks for this @laszlof! 🚀

PR #144, opened 2017-08-26
Add power state and vm state to server model

casperboone:
This PR adds the power state and the vm state of a server to the server model. Both are present in the API response (https://developer.openstack.org/api-ref/compute/#show-server-details) but currently not represented in the model.


haphan:
thank you @laszlof  and @casperboone 🚀

PR #145, opened 2017-08-28
Add server console output

casperboone:
This PR adds the ability to get the console output of a server (see [OpenStack docs](https://developer.openstack.org/api-ref/compute/#show-console-output-os-getconsoleoutput-action) ).

### DIFF
--- a/src/Compute/v2/Models/Server.php
+++ b/src/Compute/v2/Models/Server.php
@@ -247,6 +247,22 @@ class Server extends OperatorResource implements
     }
 
     /**
+     * Gets the console output of the server.
+     *
+     * @param int $length The number of lines, by default all lines will be returned.
+     * @return string
+     */
+    public function getConsoleOutput(int $length = -1)
+    {
+        $response = $this->execute($this->api->getConsoleOutput(), [
+            'id' => $this->id,
+            'length' => $length
+        ]);
+
+        return Utils::jsonDecode($response)['output'];
+    }
+
+    /**
      * Gets a VNC console for a server.
      *
      * @param  string $type The type of VNC console: novnc|xvpvnc.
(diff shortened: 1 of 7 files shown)


src/Compute/v2/Models/Server.php
```diff
@@ -239,6 +239,22 @@ public function revertResize()
     }
 
     /**
+     * Gets the console output of the server.
+     *
+     * @param int $length The number of lines, by default all lines will be returned.
+     * @return string
+     */
+    public function getConsoleOutput(int $length = -1)
```
laszlof: Could use a return typehint

casperboone: Thanks, done.

src/Compute/v2/Models/Server.php
```diff
@@ -239,6 +239,22 @@ public function revertResize()
     }
 
     /**
+     * Gets the console output of the server.
+     *
+     * @param int $length The number of lines, by default all lines will be returned.
+     * @return string
+     */
+    public function getConsoleOutput(int $length = -1): string
```
haphan: What is the logic behind default value`length = -1`?

haphan:
@casperboone Only have one small comment for you. Otherwise all looks good.

src/Compute/v2/Models/Server.php
haphan: > When given a length of -1, the API will return all lines

This is an assumption thus subject to change in the future; I think we better follow official docs that `length` shall not be included in the request if we want API to return all lines.

casperboone: Good point, will make that change soon 😄.

casperboone:
@haphan Updated the implementation, could you please have another look? 😄

haphan:
@casperboone Can you rebase with master branch?

casperboone:
@haphan Done! :)

PR #146, opened 2017-08-28
Server interface attachments

casperboone:
This PR adds three new `Server` methods to:
* Create/attach interfaces
* Detach interfaces
* Get a single interface

### DIFF
--- a/src/Compute/v2/Api.php
+++ b/src/Compute/v2/Api.php
@@ -471,6 +471,22 @@ class Api extends AbstractApi
         ];
     }
 
+    public function postInterfaceAttachment(): array
+    {
+        return [
+            'method'  => 'POST',
+            'path'    => 'servers/{id}/os-interface',
+            'jsonKey' => 'interfaceAttachment',
+            'params'  => [
+                'id'  => $this->params->urlId('server'),
+                'portId' => $this->notRequired($this->params->portId()),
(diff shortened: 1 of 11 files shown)


src/Compute/v2/Models/Server.php
```diff
@@ -329,6 +329,51 @@ public function listInterfaceAttachments(array $options = []): \Generator
     }
 
     /**
+     * Gets an interface attachment.
+     *
+     * @param string $portId The unique ID of the port.
+     * @return InterfaceAttachment
+     */
+    public function getInterfaceAttachment($portId): InterfaceAttachment
```
jamiehannaford: needs a type hint: `string $portId`

src/Compute/v2/Api.php
```diff
@@ -471,6 +471,46 @@ public function getInterfaceAttachments(): array
         ];
     }
 
+    public function getInterfaceAttachment(): array
+    {
+        return [
+            'method'  => 'GET',
+            'path'    => 'servers/{id}/os-interface/{portId}',
+            'params'  => [
+                'id'  => $this->params->urlId('server'),
```
jamiehannaford: nit: can we align array keys

casperboone:
@jamiehannaford Fixed! 😄

PR #147, opened 2017-08-30
Suspend resume server

casmit:
Add suspend/resume functions to server


casmit:
php-opencloud/openstack#152

PR #148, opened 2017-09-04
Handle Datetime serialization

tchiotludo:
The serialize method don't handle correctly `\DateTimeImmutable`

The unserialize method expect a string (\DateTimeImmutable constructor) 
```php
        } elseif (strcasecmp($type, '\datetimeimmutable') === 0) {
            $val = new \DateTimeImmutable($val);
```
But the serialize method serialize '\DateTimeImmutable' as object 
This throw an `DateTimeImmutable::__construct() expects parameter 1 to be string, object given` on calling `->populateFromArray((array) $serialize) `

### DIFF
--- a/src/Common/Resource/AbstractResource.php
+++ b/src/Common/Resource/AbstractResource.php
@@ -178,7 +178,13 @@ abstract class AbstractResource implements ResourceInterface, Serializable
             $val = $this->{$name};
 
             $fn = function ($val) {
-                return ($val instanceof Serializable) ? $val->serialize() : $val;
+                if ($val instanceof Serializable) {
+                    return $val->serialize();
+                } elseif ($val instanceof \DateTimeImmutable) {
+                    return $val->format('c');
+                } else {
+                    return $val;
+                }
             };
 
             if (is_array($val)) {


tchiotludo:
Any update ?

src/Common/Resource/AbstractResource.php
```diff
@@ -178,7 +178,13 @@ public function serialize(): \stdClass
             $val = $this->{$name};
 
             $fn = function ($val) {
-                return ($val instanceof Serializable) ? $val->serialize() : $val;
+                if ($val instanceof Serializable) {
+                    return $val->serialize();
+                } else if ($val instanceof \DateTimeImmutable) {
+                    return $val->format('Y-m-d H:i:s.u');
```
haphan: use ISO 8601 instead of hard-code date time format `$val->format('c')`

tchiotludo:
done & thanks :+1:

haphan:
Please fix travis error (PSR2) @tchiotludo  then we are good to go.

haphan:
thanks @tchiotludo

tchiotludo:
np @haphan, good to see my fork disappear :smile:

haphan:
@tchiotludo  As we are trying to maintain branch 2.x till end of March 2018, can you also back-port this fix to branch 2.0

Thanks.

PR #153, opened 2017-09-13
Add fault property to Server

casperboone:
This PR adds a fault property to the `Server` model for asynchronous errors.

### DIFF
--- a/src/Compute/v2/Models/Fault.php
+++ b/src/Compute/v2/Models/Fault.php
@@ -1,0 +1,23 @@
+<?php declare(strict_types=1);
+
+namespace OpenStack\Compute\v2\Models;
+
+use OpenStack\Common\Resource\AbstractResource;
+
+/**
+ * Represents a Compute v2 Fault.
+ */
+class Fault extends AbstractResource
+{
+    /** @var int **/
+    public $code;
+
+    /** @var \DateTimeImmutable **/
+    public $created;
+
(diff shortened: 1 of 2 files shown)


jamiehannaford:
So looking at the docs, it seems like a sample response would be this:

```json
{
  "server": {
    "fault": {
      "code" : 500,
      "created": "2010-08-10T11:59:59Z",
      "message": "No valid host was found. There are not enough hosts available.",
      "details": "..."
    }
}
```

if we return as an array, we're effectively handing over a blob of unstructured data to the user with no guaranteed contract. I'm wondering whether it's worth making a new Fault model to wrap these fields (and also converting `created` to a Datetime object)? This might be overkill though.

@haphan What do you think?

haphan:
It looks like `fault` structure is used in many scenarios as described [here](https://developer.openstack.org/api-guide/compute/faults.html), I think we should wrap it as a model @jamiehannaford

casperboone:
@jamiehannaford @haphan I've created a `Fault` model. It's extending `AbstractResource` which is not entirely fitting, but I think it's the easiest way to let the model have the behavior it needs.

jamiehannaford:
You'll need to rebase to get the travis.yml fix

casperboone:
@jamiehannaford Done! :)

haphan:
@casperboone  🚀

PR #154, opened 2017-09-19
solve issue 91

thiagodamas:
fix 2 syntax error and fix the issue when the 'region' entry under catalog its '*' (instead of a region name)

### DIFF
--- a/src/Identity/v3/Models/Endpoint.php
+++ b/src/Identity/v3/Models/Endpoint.php
@@ -77,6 +77,8 @@ class Endpoint extends OperatorResource implements Creatable, Updateable, Deleta
 
     public function regionMatches(string $value): bool
     {
+        if($this->region == '*')
+            return true;
         return $this->region && $this->region == $value;
     }
 
--- a/src/Identity/v3/Models/Service.php
+++ b/src/Identity/v3/Models/Service.php
@@ -79,7 +79,7 @@ class Service extends OperatorResource implements Creatable, Listable, Retrievab
 
     private function typeMatches(string $value): bool
     {
-        return $this->type && $this->type = $value;
+        return $this->type && $this->type == $value;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


haphan:
@thiagodamas Can you point me to an OpenStack documentation for using asterisk `*` as a region?

What does it imply in this case?

thiagodamas:
Hi, @haphan , I need to use * when using openstack from telefonica:
(get catalog)
        "endpoints": [
          {
            "url": "https://iam.sa-brazil-1.telefonicaopencloud.com/v3",
            "interface": "public",
            "region": "*",
            "region_id": "*",
            "id": "XXXXXXXXXXXXXXXXXX"
          }
        ],
        "type": "identity",
        "id": "18049c8bfa6e403793e00516d332355a",
        "name": "keystone"
      }
    ],

haphan:
@thiagodamas  I see. I have not seen any cloud provider used `*` as region id.

My concern is I don't want to tweak the SDK to support one single provider. 

Have you encounter something similar @jamiehannaford ?

jamiehannaford:
@thiagodamas @haphan I've never seen this before. Please ask Telefonica to provide a link to the docs for where this is documented. We won't merge any changes that are provider-specific, they need to be standardised for the community.

thiagodamas:
Interestingly, https://support.telefonicaopencloud.com/ and https://support.telefonicaopencloud.com/en-us/doc/pdf/20170829/20170829191636_81938.pdf (p. 64) doesnt show anything about using '*' as region "mask"

haphan:
This could be the Federated Keystone 

https://docs.openstack.org/security-guide/identity/federated-keystone.html

haphan:
One interesting observation is that since 3.2, `region` is deprecated and has been replaced with `region_id`

```
region_id  (Since v3.2) The ID of the region that contains the service endpoint.
region  (Deprecated in v3.2) The geographic location of the service endpoint.
```

Question @thiagodamas, what region do you specify in your authentication `$params`?

Issue #158, opened 2017-10-02
Opcache issues


ischenko:
Good day!

Enabling OpCache breaks the library. It turns out that part of the code relies on docblocks.
https://github.com/php-opencloud/openstack/blob/5027caab5271ab5cc7370242c827b2847ef8d2c5/src/Common/Resource/AbstractResource.php#L120-L131
As a temporary solution we had to enable `opcache.save_comments=1` on production servers (which is not good).

Is it possible to refactor the code and get rid of usage of docblocks in the library?

haphan:
> we had to enable opcache.save_comments=1 on production servers (which is not good).

Why not good @ischenko ? I don't see any benefit of doing so except saving a few bytes of the compiled opcode. There is a good reason why the default value is `1`.

Official docs:
> Disabling this configuration directive may break applications and frameworks that rely on comment parsing for annotations, including Doctrine, Zend Framework 2 and PHPUnit.

ischenko:
> Doctrine, Zend Framework 2 and PHPUnit.

If I'm not wrong Doctrine, Zend Framework 2 and PHPUnit don't enforce you to enable save_comments and can work without it. PHPUnit is not even designed to be used in production environment.

> I don't see any benefit of doing so except saving a few bytes of the compiled opcode

Comments are strings and stored as a strings. Our project has over 17k of php files. Average comment block is about 120 bytes long (some blocks are really big), assume that half of files are tests and other unused files in a production environment. Roughly we have 1MB of wasted memory I would prefer to spend it to more useful stuff.

I don't want to continue a Holy War about using of annotations, so in returning to original request: Is it possible to refactor the code and get rid of usage of docblocks in the library?

haphan:
> Is it possible to refactor the code and get rid of usage of docblocks in the library?

No, we won't spend effort trying to optimize your 1MB of memory.

jamiehannaford:
I agree with @haphan.

@ischenko While I understand your concern, saving 1MB of memory does not represent enough of a reason for us to refactor a core part of the library.

ischenko:
@jamiehannaford please see  the branch https://github.com/ischenko/openstack/commits/nodockblocks there are two commits.

 - 1a9465772852ab507b243ecc9b7aecab838b60ca  contains changes that remove usage of dockblocks
[...]

PR #159, opened 2017-10-04
Removed usage of docblocks, fixed several parsing issues

ischenko:
see #158

### DIFF
--- a/src/BlockStorage/v2/Models/Snapshot.php
+++ b/src/BlockStorage/v2/Models/Snapshot.php
@@ -2,6 +2,7 @@
 
 namespace OpenStack\BlockStorage\v2\Models;
 
+use OpenStack\Common\Resource\Alias;
 use OpenStack\Common\Resource\OperatorResource;
 use OpenStack\Common\Resource\Creatable;
 use OpenStack\Common\Resource\Deletable;
@@ -49,9 +50,18 @@ class Snapshot extends OperatorResource implements Listable, Creatable, Updateab
     protected $markerKey = 'id';
 
     protected $aliases = [
-        'created_at' => 'createdAt',
         'volume_id'  => 'volumeId',
     ];
+
+    /**
+     * @inheritdoc
+     */
+    protected function getAliases(): array
+    {
+        return parent::getAliases() + [
+            'created_at' => new Alias('createdAt', \DateTimeImmutable::class)
+        ];
+    }
 
     public function populateFromResponse(ResponseInterface $response): self
     {
--- a/src/Common/Resource/Alias.php
+++ b/src/Common/Resource/Alias.php
@@ -1,0 +1,59 @@
+<?php declare(strict_types=1);
+
+namespace OpenStack\Common\Resource;
+
+/**
+ * @package OpenStack\Common\Resource
+ */
+class Alias
+{
+    /**
+     * @var string
+     */
+    public $propertyName;
+
+    /**
+     * @var bool
+     */
+    private $isList;
+
+    /**
+     * @var string
+     */
+    private $className;
+
+    /**
+     * @param string      $propertyName A name of the property in target resource class
+     * @param string|null $className    A class name for the property value
+     * @param bool        $list         Whether value of the property should be treated as a list or not
+     */
+    public function __construct(string $propertyName, string $className = null, bool $list = false)
+    {
+        $this->isList = $list;
+        $this->propertyName = $propertyName;
+        $this->className = $className && class_exists($className) ? $className : null;
+    }
+
+    /**
+     * @param ResourceInterface $resource
+     * @param mixed             $value
+     *
+     * @return mixed
+     */
+    public function getValue(ResourceInterface $resource, $value)
+    {
+        if ($value === null || !$this->className) {
+            return $value;
+        } elseif ($this->isList && is_array($value)) {
+            $array = [];
+            foreach ($value as $subVal) {
+                $array[] = $resource->model($this->className, $subVal);
+            }
+            return $array;
+        } elseif ($this->className === \DateTimeImmutable::class) {
+            return new \DateTimeImmutable($value);
+        }
+
+        return $resource->model($this->className, $value);
+    }
+}
--- a/src/Compute/v2/Models/Image.php
+++ b/src/Compute/v2/Models/Image.php
@@ -2,6 +2,7 @@
 
 namespace OpenStack\Compute\v2\Models;
 
+use OpenStack\Common\Resource\Alias;
 use OpenStack\Common\Resource\OperatorResource;
 use OpenStack\Common\Resource\Deletable;
 use OpenStack\Common\Resource\HasMetadata;
@@ -49,6 +50,17 @@ class Image extends OperatorResource implements Listable, Retrievable, Deletable
 
     protected $resourceKey = 'image';
     protected $resourcesKey = 'images';
+
+    /**
+     * @inheritdoc
+     */
+    protected function getAliases(): array
+    {
+        return parent::getAliases() + [
+            'created' => new Alias('created', \DateTimeImmutable::class),
+            'updated' => new Alias('updated', \DateTimeImmutable::class)
+        ];
+    }
 
     /**
      * {@inheritDoc}
--- a/src/Compute/v2/Models/Server.php
+++ b/src/Compute/v2/Models/Server.php
@@ -2,6 +2,7 @@
 
 namespace OpenStack\Compute\v2\Models;
 
+use OpenStack\Common\Resource\Alias;
 use OpenStack\Common\Resource\HasWaiterTrait;
 use OpenStack\Common\Resource\Creatable;
 use OpenStack\Common\Resource\Deletable;
@@ -110,6 +111,19 @@ class Server extends OperatorResource implements
         'OS-EXT-STS:vm_state'                 => 'vmState',
         'OS-EXT-SRV-ATTR:hypervisor_hostname' => 'hypervisorHostname',
     ];
+
+    /**
+     * @inheritdoc
+     */
+    protected function getAliases(): array
+    {
+        return parent::getAliases() + [
+            'image'   => new Alias('image', Image::class),
+            'flavor'  => new Alias('flavor', Flavor::class),
+            'created' => new Alias('created', \DateTimeImmutable::class),
+            'updated' => new Alias('updated', \DateTimeImmutable::class)
+        ];
+    }
 
     /**
      * {@inheritDoc}
--- a/src/Identity/v2/Models/Catalog.php
+++ b/src/Identity/v2/Models/Catalog.php
@@ -2,6 +2,7 @@
 
 namespace OpenStack\Identity\v2\Models;
 
+use OpenStack\Common\Resource\Alias;
 use OpenStack\Common\Resource\OperatorResource;
 use OpenStack\Common\Transport\Utils;
 use Psr\Http\Message\ResponseInterface;
@@ -21,6 +22,16 @@ class Catalog extends OperatorResource implements \OpenStack\Common\Auth\Catalog
      * @var []Entry
      */
     public $entries = [];
+
+    /**
+     * @inheritdoc
+     */
+    protected function getAliases(): array
+    {
+        return parent::getAliases() + [
+            'entries' => new Alias('entries', Entry::class, true)
+        ];
+    }
 
     /**
      * {@inheritDoc}
--- a/src/Images/v2/Models/Image.php
+++ b/src/Images/v2/Models/Image.php
@@ -4,6 +4,7 @@ namespace OpenStack\Images\v2\Models;
 
 use function GuzzleHttp\Psr7\uri_for;
 use OpenStack\Common\JsonSchema\Schema;
+use OpenStack\Common\Resource\Alias;
 use OpenStack\Common\Resource\OperatorResource;
 use OpenStack\Common\Resource\Creatable;
 use OpenStack\Common\Resource\Deletable;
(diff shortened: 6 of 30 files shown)


ischenko:
Hi @jamiehannaford , @haphan 

Any comments/news on this PR?

jamiehannaford:
@ischenko Once tests pass we can take a look

ischenko:
@jamiehannaford please fix CI server: [Job #321.1](https://travis-ci.org/php-opencloud/openstack/jobs/283329505) :)

jamiehannaford:
@haphan Any idea on why CI is failing?

haphan:
Github does not allow cloning public repo without being a valid user I think.

I was able to reproduce the issue locally.

@ischenko Can you change `.travis.yml` from `composer install --prefer-source` to just `composer install`?

```
root@hpbookpro:/tmp# git clone git@github.com:php-opencloud/Sami.git
Cloning into 'Sami'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

haphan:
Github does not allow cloning public repo without being a valid user I think.

I was able to reproduce the issue locally.

@ischenko Can you change `.travis.yml` from `composer install --prefer-source` to just `composer install`?

```
root@hpbookpro:/tmp# git clone git@github.com:php-opencloud/Sami.git
Cloning into 'Sami'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

ischenko:
Done. But I think other PRs have the same issue.

ischenko:
@haphan still broken

haphan:
This has something to do with the recent change from `precise` to `trusty` in TravisCI. 

Can you add `dist: precise` in `.travis.yml` and try again @ischenko ?

ischenko:
@haphan thanks!

src/BlockStorage/v2/Models/Snapshot.php
```diff
@@ -49,7 +49,7 @@ class Snapshot extends OperatorResource implements Listable, Creatable, Updateab
     protected $markerKey = 'id';
 
     protected $aliases = [
-        'created_at' => 'createdAt',
+        'created_at' => ['createdAt', \DateTimeImmutable::class],
```
jamiehannaford: so I'm thinking of the best way we can both represent the data and abstract the parsing functionality.

instead of using a flat array (which imparts no structure), how about we create a new `Common\Resource\Alias` class? we could then move all of this alias logic inside that class and clean up `common\Resource\AbstractResource` a bit.

ischenko: If I understood correctly, `protected $aliases` will be an instance of `Common\Resource\Alias` object.

In this case initialization method should be added where alias object could be configured. Ideally if Alias object would be a static member, I don't see any reason initialize it upon each object construction.

As a variant `AbstractResource` can contain protected static method `getAlias` that should return an instance of `Common\Resource\Alias` class (here an object can be static private member, or configure parent's object, or whatever you want). This method will be used then inside of `AbstractResource::populateFromArray` method (e.g. `static::getAlias()->setProperty($this, $key, $value)`).

jamiehannaford: I was thinking more like:

```php
protected $aliases = [
   'created_at' => new Alias('createdAt', \DateTimeImmutable::class),
   'foo'        => new Alias('foo'),
];
```

```php
public function populateFromArray(array $array): self
{
   foreach ($array as $key => $val) {
      if isset($this->aliases[$key]) {
         $alias = $this->aliases[$key];
         $key = $alias->getKey();
         $val = $alias->getVal($val);
      }
      $this->$key = $val;
   }
}
```

and move all the casting logic to `Alias::getVal`.

ischenko:
@jamiehannaford I've added a commit with example of implementation of `Common\Resource\Alias` class.

There are a lot of aliases that only point to camelized property name, these aliases can also be processed inside of `Common\Resource\Alias`. For example, if there is no alias try camelize property name and if it does not exist try original key.

src/BlockStorage/v2/Models/Snapshot.php
ischenko: There are two problems:
```php 
protected $aliases = [
   'created_at' => new Alias('createdAt', \DateTimeImmutable::class),
   'foo'        => new Alias('foo'),
];
```
this won't work, expressions are not allowed for default values as far as I know. Or did I miss something?

But even so this is additional object constructions (that can be avoided) upon each `populateFromArray` call.

jamiehannaford:
@ischenko I like the idea of default camelization but I think we might be able to DRY things out a bit. For example, I don't see the need in specifying a static method for every resource that has aliases - a property should suffice IMO. Plus your Alias class is more of an Aliases wrapper. Wouldn't it be more concise for a resource to define a set of Alias objects and have the AbstractResource iterate through them?

ischenko:
@jamiehannaford you cannot instantiate an object as default value for property, so you have to provide some method where Alias objects will be created (similar to `getAlias` may be static or not) or define aliases as plain array (similar to first commit in the PR).

> For example, I don't see the need in specifying a static method for every resource that has aliases - a property should suffice IMO. Plus your Alias class is more of an Aliases wrapper.

It's just an example and can be reverted at any time :)

Here is alternate solution: return to use of `$aliases` property but also add some method (e.g. `getAliases` maybe even public) to `AbstractResource` class. The method will do only one thing: iterate thhrough $aliases and instantiate `Alias` objects. Resource classes can define their aliases in a property like before, but end class can override `getAliases` if there is need to define typed alias. It will look something like the following:
```php
class AbstractResource 
{
    // ...... here goes other code
    public function populateFromArray(array $array): self
    {
           $aliases = $this->getAliases();
           foreach ($array as $key => $val) {
               if (isset($aliases[$key]) {
                   $alias = $aliases[$key];
                   $key = $alias->getKey();
                   $val = $alias->getVal($val);
               }
               $this->$key = $val;
           }
    }
    // ...... here goes other code
    public function getAliases()
    {
        $aliases = [];
        foreach ($aliases as $alias => $property) {
            $aliases[$alias] = new Alias($property);
        }
        return $aliases;
    }
    // ...... here goes other code
}

class Resource
{
    // ...... here goes other code
    protected $aliases = ['alias' => 'propertyName'];
    // ...... here goes other code
    public function getAliases()
    {
        $aliases = parent::getAliases();
        $aliases['typedAlias'] = new Alias('createdAt', \DateTimeImmutable::class);
        return $aliases;
    }
    // ...... here goes other code
}
```
And maybe some caching at `AbstractResource` level :)

What do you think?

ischenko:
@jamiehannaford I've added an implementation of above alternative

ischenko:
@jamiehannaford the property `$aliases` can be totally replaced by method `getAliases()` or `aliases()`, for example:
```php
 protected function getAliases()
    {
        return [
            'alias' => new Alias('property'),
            'typedAlias' => new Alias('property', \DateTime::class),
        ];
    }
```
But I think that using flat array is the fastest and simplest solution. Especially that aliases are static data for internal-only use.

ischenko:
@jamiehannaford any comments?

jamiehannaford:
@ischenko Sorry, I've been busy with other things lately. I'm also not very involved with the day-to-day running of this codebase anymore (@haphan has been doing a great job).

So it seems like we have two options:

1. use flat arrays to store additional state, such as class types
2. use a method that allows us to structure aliases in their own classes

I agree that #1 is probably more concise (and maybe more performant), but I do like using a nice structure where possible because it allows new and fly-by contributors to easily grok behaviour. If I see `["foo" => ["fooBar", FooBar::class]]` it's not immediately clear to me what that does. I'd have to check the implementation. 

How about using traits to encapsulate a generic `getAliases` method and make each Resource that has aliases implement that interface? So the trait could be something like:

```php
protected $aliases = [];

protected function getAliases(): array
{
   $aliases = [];
   foreach ($this->aliases as $alias) {
      $aliases[] = new Alias($alias);
   }
   return $aliases;
}
```

and for resources that need custom logic:

```php
class FooResource implements HasAliases 
{
    uses Aliases;

    protected $aliases = ["foo" => "somethingElse"];

    protected function getAliases()
    {
        $aliases = parent::getAliases();
        $aliases["bar"] = new Alias("barKey", Bar::class);
        return $aliases;
     }
}
```

ischenko:
@jamiehannaford this looks good but I think it would be better to implement interface at AbstractResource class and override `getAliases` method only where typed alias is needed. In other cases adding aliases into `aliases` property will more than enough.

@haphan what do you think?

P.S. I also think that interface should be renamed, for example, AliasesProviderInterface or something like that.

ischenko:
Hi @jamiehannaford ,

I've updated implementation according to last comments. Please check it, if it is really need to move `getAliases` method to trait let me know. But I think it is better to leave it at `AbstractResource` class level.

ischenko:
Hi @jamiehannaford , @haphan 

Any news on this PR?

.travis.yml
```diff
@@ -21,7 +23,7 @@ before_install:
   - composer self-update
 
 before_script:
-  - composer install --prefer-source
+  - composer install
```
jamiehannaford: needs a rebase

src/Common/Resource/Alias.php
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @package OpenStack\Common\Resource
+ */
+class Alias
+{
+    /**
+     * @var bool
+     */
+    private $_isList;
```
jamiehannaford: please remove underscores

src/Common/Resource/Alias.php
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param string      $propertyName
+     * @param string|null $className
+     * @param bool        $list
+     */
+    public function __construct(string $propertyName, string $className = null, bool $list = false)
+    {
+        $this->_isList = $list;
+        $this->_propertyName = $propertyName;
+        $this->_className = (($className && class_exists($className)) ? $className : null);
```
jamiehannaford: no need for the outer brace

src/Common/Resource/Alias.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @var string
+     */
+    private $_propertyName;
+
+    /**
+     * Alias constructor.
+     *
+     * @param string      $propertyName
```
jamiehannaford: we need to document these args

src/Common/Resource/Alias.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        $this->_isList = $list;
+        $this->_propertyName = $propertyName;
+        $this->_className = (($className && class_exists($className)) ? $className : null);
+    }
+
+    /**
+     * @return string
+     */
+    public function getName()
```
jamiehannaford: let's remove this and just make the property public

src/Common/Resource/Alias.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @param ResourceInterface $resource
+     * @param mixed             $value
+     *
+     * @return mixed
+     */
+    public function getValue(ResourceInterface $resource, $value)
+    {
+        $class = $this->_className;
```
jamiehannaford: we don't really need this new var

src/Common/Resource/Alias.php
```diff
@@ -0,0 +1,71 @@
...
+
+        if ($value === null || !$class) {
+            return $value;
+        } elseif ($this->_isList && is_array($value)) {
```
jamiehannaford: let's change this to `is_iterable`

src/Compute/v2/Models/Image.php
```diff
@@ -47,10 +48,26 @@ class Image extends OperatorResource implements Listable, Retrievable, Deletable
     /** @var \DateTimeImmutable */
     public $updated;
 
+    /**
+     * @var array
+     */
+    protected $aliases = [];
```
jamiehannaford: do we need this?

src/Compute/v2/Models/Server.php
```diff
@@ -109,6 +110,19 @@ class Server extends OperatorResource implements
     ];
 
     /**
+     * @inheritdoc
+     */
+    protected function getAliases(): array
+    {
+        $aliases = parent::getAliases();
```
jamiehannaford: nit: maybe it's more succinct to do

```php
return parent::getAliases() + [
   'image' => new Alias('image', Image::class),
];
```

src/Identity/v2/Models/Catalog.php
```diff
@@ -23,6 +24,21 @@ class Catalog extends OperatorResource implements \OpenStack\Common\Auth\Catalog
     public $entries = [];
 
     /**
+     * @var array
+     */
+    protected $aliases = [];
```
jamiehannaford: is this needed?

src/Images/v2/Models/Image.php
```diff
@@ -129,12 +141,18 @@ public function update(array $data)
         // retrieve latest state so we can accurately produce a diff
         $this->retrieve();
 
-        $schema = $this->getSchema();
-        $data   = (object) $data;
+        $schema  = $this->getSchema();
+        $data    = (object) $data;
+        $aliases = array_map(
```
jamiehannaford: nit: maybe `aliasNames` is more accurate

src/Metric/v1/Gnocchi/Models/Metric.php
```diff
@@ -39,6 +40,16 @@ class Metric extends OperatorResource implements Retrievable
     ];
 
     /**
+     * @inheritdoc
+     */
+    protected function getAliases(): array
+    {
+        $aliases = parent::getAliases();
+        $aliases['resource'] = new Alias('resource', \OpenStack\Metric\v1\Gnocchi\Models\Resource::class);
```
jamiehannaford: can we import this under an alias?

src/Common/Resource/Alias.php
ischenko: `is_iterable` available from PHP 7.1 but minimal required version for the project is 7.0

ischenko:
Hi @jamiehannaford 

Requested changes (except `is_iterable`) have been pushed.

jamiehannaford:
Great work @ischenko! 🎉

PR #167, opened 2017-10-28
Implemented Policy update

samuel4x4:
Implemented Policy update

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "php-opencloud/openstack",
+    "name": "samuel4x4/openstack",
     "description": "PHP SDK for OpenStack APIs. Supports BlockStorage, Compute, Identity, Images, Networking and Metric Gnocchi",
     "keywords": [
         "php",
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "samuel4x4/openstack",
+    "name": "php-opencloud/openstack",
     "description": "PHP SDK for OpenStack APIs. Supports BlockStorage, Compute, Identity, Images, Networking and Metric Gnocchi",
     "keywords": [
         "php",
--- a/src/Common/Api/OperatorTrait.php
+++ b/src/Common/Api/OperatorTrait.php
@@ -3,7 +3,6 @@
 namespace OpenStack\Common\Api;
 
 use GuzzleHttp\Promise\PromiseInterface;
-use GuzzleHttp\Psr7\Uri;
 use function GuzzleHttp\uri_template;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Promise\Promise;
@@ -106,8 +105,6 @@ trait OperatorTrait
      */
     protected function sendRequest(Operation $operation, array $userValues = [], bool $async = false)
     {
-        $operation->validate($userValues);
-
         $options = (new RequestSerializer)->serializeOptions($operation, $userValues);
         $method = $async ? 'requestAsync' : 'request';
         $uri = uri_template($operation->getPath(), $userValues);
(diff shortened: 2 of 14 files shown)


samuel4x4:
@jamiehannaford, @haphan: I have no idea why the CI check failed, can you help me please? The only change I did in composer.json was to put temporary, in my form, the package name as "samuel4x4/openstack", so I can install it with Composer and test it.

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "php-opencloud/openstack",
+    "name": "samuel4x4/openstack",
```
jamiehannaford: this shouldn't be here

src/Common/Api/OperatorTrait.php
```diff
@@ -106,8 +105,6 @@ public function getOperation(array $definition): Operation
      */
     protected function sendRequest(Operation $operation, array $userValues = [], bool $async = false)
     {
-        $operation->validate($userValues);
```
jamiehannaford: why are we not validating?

src/Common/Resource/Updateable.php
```diff
@@ -12,7 +12,8 @@
     /**
      * Update the current resource with the configuration set out in the user options.
      *
-     * @return void
+     * @param array $userOptions
+     * @return self
      */
-    public function update();
+    public function update(array $userOptions): Updateable;
```
jamiehannaford: you're changing a public API. we cannot do this without a major version bump.

src/Identity/v3/Models/Credential.php
```diff
@@ -39,10 +39,13 @@ class Credential extends OperatorResource implements Creatable, Updateable, Retr
 
     /**
      * {@inheritDoc}
+     *
+     * @param array $data {@see \OpenStack\Identity\v3\Api::postCredentials}
      */
     public function create(array $data): Creatable
     {
         $response = $this->execute($this->api->postCredentials(), $data);
+
```
jamiehannaford: please remove

src/Identity/v3/Models/Credential.php
```diff
@@ -52,15 +55,19 @@ public function create(array $data): Creatable
     public function retrieve()
     {
         $response = $this->executeWithState($this->api->getCredential());
+
```
jamiehannaford: please remove

src/Identity/v3/Models/Credential.php
```diff
@@ -52,15 +55,19 @@ public function create(array $data): Creatable
...
     /**
      * {@inheritDoc}
+     *
+     * @param array $data {@see \OpenStack\Identity\v3\Api::patchCredential}
      */
-    public function update()
+    public function update(array $data): Updateable
     {
         $response = $this->executeWithState($this->api->patchCredential());
+
```
jamiehannaford: please remove here and all other places

src/Identity/v3/Service.php
```diff
@@ -423,6 +423,21 @@ public function createPolicy(array $options): Models\Policy
...
 
     /**
+     * Updates a policy object according to the provided options.
+     *
+     * @param string $id
+     * @param array $options {@see \OpenStack\Identity\v3\Api::patchPolicy}
+     *
+     * @return Models\Policy
+     */
+    public function updatePolicy(string $id, array $options): Models\Policy
```
jamiehannaford: this is not how operations are done. the user retrieves a model then performs the update on that object.

composer.json
samuel4x4: you're right

src/Common/Api/OperatorTrait.php
samuel4x4: it was removed by mistake

samuel4x4:
@jamiehannaford thank you for your review. I was wrong, I didn't notice there's already an existing possibility to update a Policy. Even that, I rushed and made mistakes. I'll close the PR and I will gladly use the existing feature.

PR #175, opened 2017-11-20
Auth token caching

haphan:
Proof-of-concept re-use authentication token

Usage:

Export token
```php
<?php
// Export authentication credential
$params = [];
$openstack = new OpenStack\OpenStack($params);

$token = $openstack->identityV3()->generateToken($params);

 // This is an serializable array can be re-use
$cachedToken = $token->export();

// Save token to file
file_put_contents('cache.json', \json_encode($cachedToken));

// Or even better wrt scalability, store token in memcache
$memcache->set('array', $cachedToken, $expiration);

```

Use exported token
```php
// Authenticate using cache credential
$cachedToken = \json_decode(file_get_contents('cache.json'), true);

// params need to have all other mandatory fields such as `authUrl`, `user`, etc..
$params = [
  'cachedToken' => $cachedToken
]; 

$openstack = new OpenStack\OpenStack($params);
```

Design concerns
- Token expiration is not validated. Possibly rely on `expires_at` flag to trigger re-auth
- Increase in memory usage since `$token` holds a copy of API response in order to export it when needed. Depends on the size of catalog, this increase could range from 5-10KB

### DIFF
--- a/doc/services/identity/v3/tokens.rst
+++ b/doc/services/identity/v3/tokens.rst
@@ -50,18 +50,17 @@ Cache authentication token
 Use case
 ~~~~~~~~
 
-Before one can start calling any API, a very first step that this SDK will do is to authenticate with Identity service
-using user's credential.
+Before the SDK performs an API call, it will first authenticate to the OpenStack Identity service using the provided
+credentials.
 
-If the user's credential is valid, Identity service responses with an authentication token embedded in X-Auth-Token
-header and services catalog. The SDK will then use this authentication token and services catalog in all subsequent API
-calls.
+If the user's credential is valid, credentials are valid, the Identity service returns an authentication token. The SDK
+will then use this authentication token and service catalog in all subsequent API calls.
 
-This setup typically works well for command line type of applications. However, for web-based applications, performance
+This setup typically works well for CLI applications. However, for web-based applications, performance
 is undesirable since authentication step adds ~100ms to the response time.
 
-In order to improve performance, SDK allows users to export and store authentication token and re-use it when it is
-still valid.
+In order to improve performance, the SDK allows users to export and store authentication tokens, and re-use until they
+expire.
 
 
 Generate token and persist to file
@@ -71,7 +70,7 @@ Generate token and persist to file
 
 
 For scalability, it is recommended that cached tokens are stored in persistent storage such as memcache or redis instead
-of local file.
+of a local file.
 
 Initialize Open Stack using cached authentication token
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/services/identity/v3/tokens.rst
+++ b/doc/services/identity/v3/tokens.rst
@@ -43,3 +43,37 @@ Revoke token
 
 .. sample:: identity/v3/tokens/revoke_token.php
 .. refdoc:: OpenStack/Identity/v3/Service.html#method_revokeToken
+
+Cache authentication token
+--------------------------
+
+Use case
+~~~~~~~~
+
+Before one can start calling any API, a very first step that this SDK will do is to authenticate with Identity service
+using user's credential.
+
+If the user's credential is valid, Identity service responses with an authentication token embedded in X-Auth-Token
+header and services catalog. The SDK will then use this authentication token and services catalog in all subsequent API
+calls.
+
+This setup typically works well for command line type of applications. However, for web-based applications, performance
+is undesirable since authentication step adds ~100ms to the response time.
+
+In order to improve performance, SDK allows users to export and store authentication token and re-use it when it is
+still valid.
+
+
+Generate token and persist to file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. sample:: identity/v3/tokens/export_authentication_token.php
+
+
+For scalability, it is recommended that cached tokens are stored in persistent storage such as memcache or redis instead
+of local file.
+
+Initialize Open Stack using cached authentication token
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. sample:: identity/v3/tokens/use_cached_authentication_token.php
--- a/samples/identity/v3/tokens/export_authentication_token.php
+++ b/samples/identity/v3/tokens/export_authentication_token.php
@@ -30,4 +30,6 @@ file_put_contents('token.json', json_encode($token->export()));
 
 // Alternatively, one may persist token to memcache or redis
 // Redis and memcache then can purge the entry when token expires.
+
+/**@var \Memcached $memcache */
 $memcache->set('token', $token->export(), $token->expires->format('U'));
--- a/samples/identity/v3/tokens/export_authentication_token.php
+++ b/samples/identity/v3/tokens/export_authentication_token.php
@@ -30,4 +30,4 @@ file_put_contents('token.json', json_encode($token->export()));
 
 // Alternatively, one may persist token to memcache or redis
 // Redis and memcache then can purge the entry when token expires.
-$memcache->set('token', $token->export, $token->expires->format('U'));
+$memcache->set('token', $token->export(), $token->expires->format('U'));
--- a/samples/identity/v3/tokens/export_authentication_token.php
+++ b/samples/identity/v3/tokens/export_authentication_token.php
@@ -1,0 +1,33 @@
+<?php
+
+require 'vendor/autoload.php';
+
+$params = [
+    'authUrl' => '{authUrl}',
+    'region'  => '{region}',
+    'user'    => [
+        'name'     => '{username}',
+        'password' => '{password}',
+        'domain'   => ['id' => '{domainId}']
+    ],
+    'scope' => [
+        'project' => ['id' => '{projectId}']
+    ]
+];
+
+$openstack = new OpenStack\OpenStack($params);
+
+$identity = $openstack->identityV3();
+
+$token = $identity->generateToken($params);
+
+// Display token expiry
+echo sprintf('Token expires at %s'. PHP_EOL, $token->expires->format('c'));
+
+// Save token to file
+file_put_contents('token.json', json_encode($token->export()));
+
+
+// Alternatively, one may persist token to memcache or redis
+// Redis and memcache then can purge the entry when token expires.
+$memcache->set('token', $token->export, $token->expires->format('U'));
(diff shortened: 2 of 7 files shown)


haphan:
Since you guys expressed interested in this feature, please do have a comment @weknowsoftware, @gillesgw 

@jamiehannaford Please comment

src/Identity/v3/Models/Token.php
```diff
@@ -43,6 +44,8 @@ class Token extends OperatorResource implements Creatable, Retrievable, \OpenSta
     protected $resourceKey = 'token';
     protected $resourcesKey = 'tokens';
 
+    protected $cacheCredential;
```
jamiehannaford: wondering what the best name for this is. `cachedToken`, `cachedResponse`, `cachedArray`?

"credential" seems quite ambiguous with ref to tokens

src/Identity/v3/Models/Token.php
```diff
@@ -116,6 +119,22 @@ public function create(array $data): Creatable
...
+        return $token;
+    }
+
+    /**
+     * Retrieves an array serializable representation of authentication token.
+     * Can be use to initialise OpenStack object using $params['cachedCredential']
+     *
+     * @return array
+     */
+    public function exportCredential(): array
```
jamiehannaford: this method seems pretty superfluous. how about we make `cachedToken` public and change this method to something like `exportToFile(string $path)` that persists to disk?

I think it's safe to make some assumption in a helper method. If people want to control their own caching implementation they can use `$token->cachedCredential` directly.

src/Identity/v3/Models/Token.php
jamiehannaford: we should also add a description saying that the array's structure is a modified version of the POST response, since it has an "id" key added

src/Identity/v3/Service.php
```diff
@@ -31,7 +31,11 @@ public function authenticate(array $options): array
     {
         $authOptions = array_intersect_key($options, $this->api->postTokens()['params']);
 
-        $token = $this->generateToken($authOptions);
+        if (!empty($options['cachedCredential'])) {
+            $token = $this->generateTokenFromCache($options['cachedCredential']);
```
jamiehannaford: imagine that I'm passing in an expired array here, should a fatal error be raised? My initial thought is yes. WDYT?

src/Identity/v3/Models/Token.php
josiasmontag: I think `exportToFile` would be to specific. Another use case of this is caching the tokens in redis, memcache etc.

jamiehannaford: Yes, but persistence to file is a common denominator scenario we can predict and offer a helper for.

haphan: > persistence to file is a common denominator scenario

Hmm hard to make this assumption here since we don't have data to support. 

I tend to agree with @josiasmontag that SDK should not handle persistent data. `12-factors` methodology encourages apps not to store data locally but in a stateful back-end service.

Applications employ this SDK should decide where to store the token; it only takes a one-liner to write to file anyway.

haphan: > how about we make `cachedToken` public 

Not a good idea.

src/Identity/v3/Service.php
haphan: Good point. Validation added.

haphan:
When you got time @jamiehannaford

src/Identity/v3/Models/Token.php
```diff
@@ -116,6 +119,30 @@ public function create(array $data): Creatable
         }
 
         $response = $this->execute($this->api->postTokens(), $data);
-        return $this->populateFromResponse($response);
+        $token = $this->populateFromResponse($response);
+
+
```
jamiehannaford: nit: remove extra line

src/Identity/v3/Models/Token.php
```diff
@@ -116,6 +119,30 @@ public function create(array $data): Creatable
...
+        /**
+         * Cache response as an array to export if needed.
+         *  Added key `id` which is auth token from HTTP header X-Subject-Token
```
jamiehannaford: nit: can we just use a multiline `//` comment for brevity

src/Identity/v3/Models/Token.php
```diff
@@ -116,6 +119,30 @@ public function create(array $data): Creatable
...
+         */
+        $this->cachedToken = Utils::flattenJson(Utils::jsonDecode($response), $this->resourceKey);
+        $this->cachedToken['id'] = $token->id;
+
+        return $token;
+    }
+
+    /**
+     * Retrieves an array serializable representation of authentication token.
```
jamiehannaford: Returns a serialized representation of an authentication token.

src/Identity/v3/Service.php
```diff
@@ -46,6 +54,19 @@ public function authenticate(array $options): array
             $type, $name, $region, $interface));
     }
 
+    /**
+     * Generates authentication token from cached token using `$token->export()`
+     *
+     * @param array $cachedToken {@see \OpenStack\Identity\v3\Models\Token::export}
+     *
+     * @return Models\Token
+     *
```
jamiehannaford: nit: remove line

haphan:
Added use case and sample scripts. When you have time @jamiehannaford

src/Identity/v3/Service.php
```diff
@@ -31,7 +31,15 @@ public function authenticate(array $options): array
     {
         $authOptions = array_intersect_key($options, $this->api->postTokens()['params']);
 
-        $token = $this->generateToken($authOptions);
+        if (!empty($options['cachedToken'])) {
+            $token = $this->generateTokenFromCache($options['cachedToken']);
+
+            if ($token->hasExpired()) {
+                throw new \RuntimeException(sprintf('Cached token has expired. You may need to re-generate a new token.'));
```
jamiehannaford: nit: since we're using `sprintf` how about output the expired date?

src/Identity/v3/Models/Token.php
```diff
@@ -116,6 +119,27 @@ public function create(array $data): Creatable
...
+        $this->cachedToken = Utils::flattenJson(Utils::jsonDecode($response), $this->resourceKey);
+        $this->cachedToken['id'] = $token->id;
+
+        return $token;
+    }
+
+    /**
+     * Returns a serialized representation of an authentication token.
+     *
+     * Initialize OpenStack object using $params['cachedToken'] to reduce HTTP authentication call.
```
jamiehannaford: to reduce the amount of HTTP calls

src/Identity/v3/Models/Token.php
```diff
@@ -116,6 +119,27 @@ public function create(array $data): Creatable
...
+     *
+     * This array is a modified version of POST response to `auth/tokens`. Do not manually modify this array.
```
jamiehannaford: This array is a modified response of the POST request to `auth/tokens`

samples/identity/v3/tokens/use_cached_authentication_token.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+require 'vendor/autoload.php';
+
+
+
```
jamiehannaford: nit: remove extra spaces

samples/identity/v3/tokens/export_authentication_token.php
```diff
@@ -0,0 +1,33 @@
...
+// Display token expiry
+echo sprintf('Token expires at %s'. PHP_EOL, $token->expires->format('c'));
+
+// Save token to file
+file_put_contents('token.json', json_encode($token->export()));
+
+
+// Alternatively, one may persist token to memcache or redis
+// Redis and memcache then can purge the entry when token expires.
+$memcache->set('token', $token->export(), $token->expires->format('U'));
```
jamiehannaford: maybe we should link to the library, or at least define what type `$memcache` is

doc/services/identity/v3/tokens.rst
```diff
@@ -43,3 +43,37 @@ Revoke token
...
 .. sample:: identity/v3/tokens/revoke_token.php
 .. refdoc:: OpenStack/Identity/v3/Service.html#method_revokeToken
+
+Cache authentication token
+--------------------------
+
+Use case
+~~~~~~~~
+
+Before one can start calling any API, a very first step that this SDK will do is to authenticate with Identity service
```
jamiehannaford: Before the SDK performs an API call, it will first authenticate to the OpenStack Identity service using the provided credentials.

doc/services/identity/v3/tokens.rst
```diff
@@ -43,3 +43,37 @@ Revoke token
...
+using user's credential.
+
+If the user's credential is valid, Identity service responses with an authentication token embedded in X-Auth-Token
```
jamiehannaford: credentials are valid, the Identity service returns an authentication token

doc/services/identity/v3/tokens.rst
```diff
@@ -43,3 +43,37 @@ Revoke token
...
+header and services catalog. The SDK will then use this authentication token and services catalog in all subsequent API
```
jamiehannaford: the service catalog ... and service catalog

doc/services/identity/v3/tokens.rst
```diff
@@ -43,3 +43,37 @@ Revoke token
...
+calls.
+
+This setup typically works well for command line type of applications. However, for web-based applications, performance
```
jamiehannaford: for CLI applications

doc/services/identity/v3/tokens.rst
```diff
@@ -43,3 +43,37 @@ Revoke token
...
+is undesirable since authentication step adds ~100ms to the response time.
+
+In order to improve performance, SDK allows users to export and store authentication token and re-use it when it is
```
jamiehannaford: the SDK ... authentication tokens,

doc/services/identity/v3/tokens.rst
```diff
@@ -43,3 +43,37 @@ Revoke token
...
+
+
+Generate token and persist to file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. sample:: identity/v3/tokens/export_authentication_token.php
+
+
+For scalability, it is recommended that cached tokens are stored in persistent storage such as memcache or redis instead
+of local file.
```
jamiehannaford: a local file

src/Identity/v3/Service.php
haphan: Good idea.

src/Identity/v3/Models/Token.php
haphan: The token is actually a `response`

jamiehannaford:
Awesome work @haphan !

PR #185, opened 2017-12-14
Fixing Policy PATCH method

samuel4x4:
Added missing `jsonKey` key on patchPolicy() method and fixed affected test.


haphan:
thanks @samuel4x4

Issue #181, opened 2017-12-06
AbstractResource must be compatible with ResourceInterface::populateFromResponse


ProPheT777:
Hello, 

Since migration to php 7.2 I got this error, with the latest tagged version : 

```
Compile Error: Declaration of OpenStack\Common\Resource\AbstractResource::populateFromResponse(Psr\Http\Message\ResponseInterface $response): OpenStack\Common\Resource\AbstractResource must be compatible with OpenStack\Common\Resource\ResourceInterface::populateFromResponse(Psr\Http\Message\ResponseInterface $response): OpenStack\Common\Resource\ResourceInterface

  at vendor/php-opencloud/openstack/src/Common/Resource/AbstractResource.php:16
```

haphan:
SDK has not been tested with 7.2 since it is relatively new.

I will add php 7.2 to travis.

ProPheT777:
Ok, nice to Know, thanks you, I'll sens PR if I have free Time to fixe that

haphan:
@ProPheT777 I just ran tests with php 7.2 and unfortunately, the SDK may require a bit of refactoring as well as breaking backward compatibility.
```
Fatal error: Cannot use OpenStack\ObjectStore\v1\Models\Object as Object because 'Object' is a special class name in /work/tests/unit/ObjectStore/v1/Models/ContainerTest.php on line 11```

phansys:
This fix has nothing to do with PHP 7.2, since the same result occurs with PHP 7.0 or 7.1: https://3v4l.org/dkLdi.

haphan:
Fix issue should be fixed in latest 3.x and 2.x version. Let me know if it is not the case.

PR #184, opened 2017-12-13
PHP 7.2 Compatibility (+ updated php-cs-fixer & PHPUnit)

mrwogu:

PR #188, opened 2017-12-21
Refactoring and fixes for PHP 7.2

noo-zh:
This solves [issue 181](https://github.com/php-opencloud/openstack/issues/181).

Fixes:

- fix incompatible declaration of `OpenStack\Common\Resource\ResourceInterface::populateFromResponse`
- rename `OpenStack\ObjectStore\v1\Models\Object` to `OpenStack\ObjectStore\v1\Models\ObjectEntity` due to new [object typehint](https://wiki.php.net/rfc/object-typehint)
- refactor `self` typehints in child classes

Because of renaming `Object -> ObjectEntity`, this should be a candidate for `v3.0.0` release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ matrix:
   include:
     - php: 7.0
     - php: 7.1
+    - php: 7.2
     - php: nightly
   allow_failures:
     - php: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
 before_script:
   - composer install --prefer-source
   - vendor/bin/parallel-lint --exclude vendor .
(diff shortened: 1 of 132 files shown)


mrwogu:
@noo-zh you didn't build it under PHP 7.2 in travis. Your repo does not work.

noo-zh:
@mrwogu I see. Travis config updated. It seems only tests may be out of date. I actually already use my branch in production and everything works well (of course, I don't use all features but still).

mrwogu:
@noo-zh please look at my PR. I've fixed everything for PHP 7.2, 7.1 and 7.0. Now there is one test failed.

haphan:
Version 3.0 should be compatible with php 7.2 now thanks to #184

PR #193, opened 2018-01-16
fix metadata header parsing

nicolasmure:
Hello,

This PR fixes the followiong metadata parsing error :
```php
$mask = 'X-Object-Meta-';
$header = 'X-Object-Meta-Manufacturer';

echo ltrim($header, $mask); // "nufacturer"
```
See live example [here](https://3v4l.org/f9aLm).

Actually, `ltrim` does not replaces by the whole string but uses chars in the given mask instead, and stops when it encounter a char which is not in the mask.

### DIFF
--- a/src/ObjectStore/v1/Models/MetadataTrait.php
+++ b/src/ObjectStore/v1/Models/MetadataTrait.php
@@ -11,9 +11,9 @@ trait MetadataTrait
         $metadata = [];
 
         foreach ($response->getHeaders() as $header => $value) {
-            $position = strpos($header, static::METADATA_PREFIX);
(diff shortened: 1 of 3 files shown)


haphan:
Nice catch @nicolasmure! Surprisingly no one has reported this issue.

Performance wise, for this simple job, do you think we can just simply use `substr` instead of regex?

```php
$name = substr($header, strlen(static::METADATA_PREFIX));
```

nicolasmure:
Sure @haphan :) I updated the code.

haphan:
thanks @nicolasmure  🎸

PR #194, opened 2018-01-19
Add tenantId to postSecurityRules api

Jahsis:
Add ability to specify tenant_id when create Security Group Rule. It is useful if requests are made from admin tenant. Instead Security Group Rule is created with callers tenant ID.


haphan:
🚀 Thanks @Jahsis

PR #195, opened 2018-01-22
Fix issue with retrieving FloatingIP

Jahsis:
Fix issue with retrieving FloatingIP data. Previously retrieved only object without populated data.

### DIFF
--- a/src/Networking/v2/Extensions/Layer3/Models/FloatingIp.php
+++ b/src/Networking/v2/Extensions/Layer3/Models/FloatingIp.php
@@ -70,8 +70,7 @@ class FloatingIp extends OperatorResource implements Listable, Creatable, Retrie
 
     public function retrieve()
     {
-        $response = $this->execute($this->api->getFloatingIp(),
-                                   ['id' => (string)$this->id]);
+        $response = $this->executeWithState($this->api->getFloatingIp());
         $this->populateFromResponse($response);
     }
 }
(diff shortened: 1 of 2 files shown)


haphan:
Can you post a reproducible code snippet? @Jahsis

src/Networking/v2/Extensions/Layer3/Models/FloatingIp.php
```diff
@@ -70,6 +70,8 @@ public function delete()
 
     public function retrieve()
     {
-        $this->executeWithState($this->api->getFloatingIp());
+        $response = $this->execute($this->api->getFloatingIp(),
```
haphan: Shouldn't this equivalent to the following?

```php
$response =  $this->executeWithState($this->api->getFloatingIp());
```

haphan:
Good catch @Jahsis.
Please check my comment and let me know if its ok.

src/Networking/v2/Extensions/Layer3/Models/FloatingIp.php
Jahsis: Yes, looks like it is same. I guided myself more by other models in this project, for example:

https://github.com/php-opencloud/openstack/blob/master/src/Networking/v2/Models/Network.php#L53
https://github.com/php-opencloud/openstack/blob/master/src/Networking/v2/Models/Port.php#L149

And so on.

Jahsis: Looks like your suggestion is much more cleaner, I will do changes.

Jahsis:
@haphan please, check my changes.

PR #214, opened 2018-02-12
Add 'subdir' as alias for 'name' in StorageObject

icewind1991:
When listing objects with `delimiter=/` the response for any sub directory will only contain `{"subdir": "folder/"}`

Without this alias there is no way get a list of sub directories (without listing *all* objects)

### DIFF
--- a/src/ObjectStore/v1/Models/StorageObject.php
+++ b/src/ObjectStore/v1/Models/StorageObject.php
@@ -47,6 +47,7 @@ class StorageObject extends OperatorResource implements Creatable, Deletable, Ha
     protected $aliases = [
         'bytes' => 'contentLength',
         'content_type' => 'contentType',
+        'subdir' => 'name',
     ];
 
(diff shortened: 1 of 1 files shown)


haphan:
PR  LGTM, however, `subdir` is not documented anywhere as far as I can see.

@icewind1991 Do you have any document explains this field? 

Ref: https://bugs.launchpad.net/swift/+bug/1505768

icewind1991:
I couldn't find any documentation about this, I figured it out when debugging an upgrade from the old php-opencloud which [does handle it](https://github.com/rackspace/php-opencloud/blob/4443dcee20fcf140bfdcf00345fc995f61f19081/lib/OpenCloud/ObjectStore/Resource/DataObject.php#L100)

haphan:
thanks @icewind1991

haphan:
This issue has been backported to branch 2.x via #218

PR #236, opened 2018-04-20
fix(identity-service): incorrect logic

JLHwung:
Copied from #154 as the `typeMatches` logic is apparently incorrect.


JLHwung:
Closed due to lack of activity.

PR #238, opened 2018-04-30
Fixed ObjectStorage PHPDoc of contentLength and lastModified

vitalybaev:
This properties have wrong types in PHPDoc


haphan:
thanks @vitalybaev 🚀

vitalybaev:
@haphan Thanks a lot! Could you ship a new patch release, cause Scrutinizer marks our code as bug due to incompatible types.

PR #250, opened 2018-08-21
Added update method to ObjectStorage

haphan:


haphan:
This change will also be incorporate into PR #251

PR #251, opened 2018-08-22
Added tempUrl generating capability

haphan:
Usage:
```php

$swiftService = $openstack->objectStoreV1();

$expire = time() + 3600;
$path = '/v1/AUTH_account/container/object';
$key = 'mys33rtkey';

$tempUrl = $swiftService->tempUrl('GET', $expire, $path, $key);
```

This method loosely follows swift command's way to generate temporary url: 
```
swift tempurl $METHOD $EXPIRE $PATH $KEY
```

Some features this PR tries to support:
- [] tempUrl for single object
- [] tempUrl for multiple objects by prefix-based path
- [] multiple hash digest algorithm (sha1, sha256, sha512)
- [] ip range restriction


src/ObjectStore/v1/Service.php
```diff
@@ -86,4 +86,42 @@ public function containerExists(string $name): bool
...
         }
     }
+
+    /**
+     * Creates a temporary URL to access object in private containers.
+     * This method loosely follows swift command's way to generate temporary url: `swift tempurl $METHOD $EXPIRE $PATH $KEY`.
+     *
+     * @param string $method  An HTTP method to allow for this temporary URL. Any of GET, POST, HEAD, PUT, POST, DELETE.
+     * @param int    $expires Unix timestamp
+     * @param string $path    The full path or storage URL to the Swift object. Example: '/v1/AUTH_account/c/o'  or: 'http://saio:8080/v1/AUTH_account/c/o'
```
MSC29: Hi there. I just re-used this code as I need to generate temporary urls.
It took me a while to notice the issue but here it is in the doc "Do not URL-encode the path when you generate the HMAC-SHA1 signature. However, when you make the actual HTTP request, you should properly URL-encode the URL.".

I'd add this to the doc to ensure an non-encoded $path is passed in.

src/ObjectStore/v1/Service.php
```diff
@@ -86,4 +86,42 @@ public function containerExists(string $name): bool
...
+        } else {
+            $message = sprintf("%s\n%s\n%s", $method, $expires, $path);
+        }
+
+        $signature = hash_hmac($digest, $message, $key);
+
+        // sha512 requires prefixing signature
+        $signature = 'sha512' === $digest ? 'sha512:'.$signature : $signature;
+
+        return sprintf('%s?temp_url_sig=%s&temp_url_expires=%s', $path, $signature, $expires);
```
MSC29: Given the $path is not encoded, I'd suggest encoding the actual temporary url path

```suggestion
        return sprintf('%s?temp_url_sig=%s&temp_url_expires=%s', urlencode($path), $signature, $expires);
```

drzraf:
In stein and subsequent releases of Swift, the hmac is [done differently](https://docs.openstack.org/swift/train/middleware.html#formpost)
```python
hmac_body = '%s\n%s\n%s\n%s\n%s' % (path, redirect, max_file_size, max_file_count, expires)
```

drzraf:
My main remark is that the function returns the `signature` (what is needed from this function) bundled with `$expires` and `$path` we just passed.
Let imagine I'm having a `Psr\Http\Message\UriInterface`  (as passed by`$container->getObject('')->getPublicUri();`), I'd rather only want the signature, in order to then call:
`$uri->withQuery(sprintf('temp_url_sig=%s&temp_url_expires=%s', $signature, $expires))->__toString();`

It's especially annoying since if `prefix` are used, the `prefix:<path>` format used to generate the hash is not suitable for actual Uri building.
I think the function should just return the `$signature` (string)

PR #254, opened 2018-08-29
Minor Updates

haphan:
Some minor updates I cherry-picked from a project I've been working on.

You have time to do a quick code review @laszlof?

### DIFF
--- a/src/BlockStorage/v2/Models/Volume.php
+++ b/src/BlockStorage/v2/Models/Volume.php
@@ -62,6 +62,9 @@ class Volume extends OperatorResource implements Creatable, Listable, Updateable
     /** @var string */
     public $host;
 
+    /** @var string */
+    public $bootable;
+
     /** @var array */
     public $metadata = [];
 
@@ -155,5 +158,22 @@ class Volume extends OperatorResource implements Creatable, Listable, Updateable
         $json = Utils::jsonDecode($response);
 
         return isset($json['metadata']) ? $json['metadata'] : [];
+    }
+
+    public function setBootable(bool $bootable)
+    {
+        $bootable = boolval($bootable);
+        $this->execute($this->api->postVolumeBootable(), ['id' => $this->id, 'bootable' => $bootable]);
+    }
+
+    public function setImageMetadata(array $metadata)
+    {
+        $this->execute($this->api->postImageMetadata(), ['id' => $this->id, 'metadata' => $metadata]);
+    }
+
+    public function resetStatus(array $options)
+    {
+        $options += ['id' => $this->id];
+        $this->execute($this->api->postResetStatus(), $options);
     }
 }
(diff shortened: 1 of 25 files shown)


laszlof:
It's currently pretty late here. I can take a look in the AM if that works?

src/BlockStorage/v2/Models/Volume.php
```diff
@@ -156,4 +159,42 @@ public function parseMetadata(ResponseInterface $response): array
 
         return isset($json['metadata']) ? $json['metadata'] : [];
     }
+
+    /**
+     * Update the bootable status for a volume, mark it as a bootable volume.
+     *
+     * @param bool $bootable
+     */
+    public function setBootable(bool $bootable)
```
laszlof: Seems like this should have a default value (true). The name of the method insinuates that it makes the volume bootable, rather than necessarily setting a flag.

src/BlockStorage/v2/Models/Volume.php
```diff
@@ -156,4 +159,42 @@ public function parseMetadata(ResponseInterface $response): array
...
+    {
+        $bootable = boolval($bootable);
```
laszlof: Casting here would be more performant. Just splitting hairs here though.

src/BlockStorage/v2/Models/Volume.php
```diff
@@ -156,4 +159,42 @@ public function parseMetadata(ResponseInterface $response): array
...
+    {
+        $this->execute($this->api->postImageMetadata(), ['id' => $this->id, 'metadata' => $metadata]);
+    }
+
+    /**
+     * Administrator only. Resets the status, attach status, and migration status for a volume. Specify the os-reset_status action in the request body.
+     *
+     * @param array $options
+     *
+     * $options['status']             = (string)    The volume status.
```
laszlof: I think the `=` here is supposed to line up with the next 2 lines.

src/BlockStorage/v2/Models/Volume.php
```diff
@@ -156,4 +159,42 @@ public function parseMetadata(ResponseInterface $response): array
...
+     * $options['migrationStatus']   = (string)    The volume migration status.
+     * $options['attachStatus']      = (string)    The volume attach status.      [OPTIONAL]
+     *
+     * @see https://developer.openstack.org/api-ref/block-storage/v2/index.html#volume-actions-volumes-action
+     */
+    public function resetStatus(array $options)
+    {
+        $options += ['id' => $this->id];
```
laszlof: Might prefer `array_merge` here. If they pass in an `id` this isnt going to overwrite it (Maybe you want that?).

src/ObjectStore/v1/Models/StorageObject.php
```diff
@@ -35,10 +35,10 @@ class StorageObject extends OperatorResource implements Creatable, Deletable, Ha
     /** @var string */
     public $contentType;
 
-    /** @var int */
+    /** @var \DateTimeImmutable */
```
laszlof: This typehint cant be right. This is a `Content-Length` header, it should probably be a string (according to API spec)

https://developer.openstack.org/api-ref/object-store/#get-object-content-and-metadata

laszlof:
Looks good pending travis

haphan:
Thank you @laszlof

PR #262, opened 2018-10-24
Replaced attachmentId with volumeId when detaching volumes as per doc…

tofsing:
…umentation: https://docs.openstack.org/nova/latest/configuration/policy.html


k0ka:
As far as I see, `volumeAttachment->id` is the same as `volume->id` (at least now) - https://docs.openstack.org/api-ref/compute/#show-a-detail-of-a-volume-attachment
This PR would break backward compatibility (in case someone is using the `API` class which is marked as `@internal`). So this PR doesn't make sense right now.

PR #271, opened 2018-12-18
Implement resuming and suspending of servers

mzur:
This implements the [resume](https://developer.openstack.org/api-ref/compute/?expanded=resume-suspended-server-resume-action-detail#resume-suspended-server-resume-action) and [suspend](https://developer.openstack.org/api-ref/compute/?expanded=suspend-server-suspend-action-detail#suspend-server-suspend-action) actions for compute instances. I did not add these actions to the docs because start and stop weren't mentioned there as well. But I can still add it if you like.

Please review and tell me if you are willing to merge. This is basically #147 which was closed and not merged for no reason that I can see.


mzur:
Some tests are failing because of #272 but these are unrelated to the proposed additions.

k0ka:
released in https://github.com/php-opencloud/openstack/releases/tag/v3.7.0

PR #272, opened 2018-12-18
Fix "Missing header delimiter" errors of new GuzzleHttp version

mzur:
A newer `guzzlehttp/psr7` version caused some unit tests to fail with `InvalidArgumentException: Invalid message: Missing header delimiter`. This PR adds the header delimiter to fix the tests.


haphan:
Thanks @mzur  Didn't notice you created this PR. Anyway I have fixed this issue and merged to master.

PR #275, opened 2019-01-15
[smarcet]

smarcet:
* added application credentials option to identity api v3 ( get token )
https://docs.openstack.org/keystone/rocky/user/application_credentials.html


mzur:
Is this the same than #336?

k0ka:
superseded by https://github.com/php-opencloud/openstack/pull/336

PR #284, opened 2019-06-27
Support volume_type in server create

obi12341:
This little change will support volume_type in server create. This will be mandatory if you have multiple volume types. Example one 'ssd' and one 'hdd' volume type.


src/Compute/v2/Params.php
```diff
@@ -356,6 +356,11 @@ public function blockDeviceMapping(): array
                         'sentAs'      => 'volume_size',
                         'description' => 'Size of the volume created if we are doing vol creation',
                     ],
+                    'volumeType' => [
+                        'type'        => self::STRING_TYPE,
+                        'sentAs'      => 'volume_type',
+                        'description' => 'This can be used to specify the type of volume which the compute service will create and attach to the server.',
```
TheDigitalOrchard: Suggest removing words "This can be used to...". Just start with "Specify the type of...".

PR #288, opened 2019-07-27
Fixes for Issue 287

TheDigitalOrchard:
Two of the three suggested changes included.

- Allow wildcard region
- Fix == comparison operator

The third change is unclear if that's a bug or not. Not including.


peter279k:
The Travis CI build is failed because the coding style check is failed during `PHP-CS-Fixer` execution.

haphan:
Accepting a wildcard here means we are creating an exception for OVH OpenStack specifically. I don't see any other provider uses this convention; and more importantly; the behaviour is not documented at all from official Openstack website. Are we going to keep doing is a real question to ask as a community

IMHO, we must, at least, in the source code and docs fully document this behaviour and the rationale. Otherwise I am OK with this PR

Issue #292, opened 2019-11-20
The body of an object is fetched by default


chapa:
Contrary to what's written in the [phpdoc of the method](https://github.com/php-opencloud/openstack/blob/3eba351cd04cf73a54984334e5da2fe4cf6efe1a/src/ObjectStore/v1/Models/StorageObject.php#L140-L141), calling `StorageObject::download()` will fetch the whole file before returning it (as a `StreamInterface`, indeed).

I don't know if it has always been the case, maybe it's a change on the Guzzle side, but now the whole file is fetched. It becomes a problem with large files because we can't start processing them (e.g. returning them to client) before they've been completely downloaded.

I found that using the [`'stream'`](http://docs.guzzlephp.org/en/stable/request-options.html#stream-option) option from Guzzle solve the issue, but it's not as simple as that...

By adding this option in the `'requestOptions'` option of `OpenStack`, it will brake the authentication request that doesn't seems to support streams. The `Utils::jsonDecode($response)` will return an empty string, so `Utils::jsonDecode($response)['access']['token']` will fail :
https://github.com/php-opencloud/openstack/blob/3eba351cd04cf73a54984334e5da2fe4cf6efe1a/src/Identity/v2/Models/Token.php#L46

The problem seems to be here, where the body of the response is casted to a string instead of properly handled :
https://github.com/php-opencloud/openstack/blob/3eba351cd04cf73a54984334e5da2fe4cf6efe1a/src/Common/Transport/Utils.php#L23

But it's not as simple as changing `(string) $response->getBody()` by `$response->getBody()->getContents()`, because the stream returned by Guzzle is not seekable (when you set the `'stream'` option to `true`). `Utils::jsonDecode($response)` is called more than one with the same `$response`, so the second time it also returns an empty string (I guess because the content of the stream has been consumed during the first call).

I managed to get around it with a dirty solution consisting of having two `OpenStack` objects : one with `'requestOptions' => ['stream' => true]` and the other without. I can then do something like this :
```php
$openstack = new OpenStack([
    // ...
]);

$streamedOpenstack = new OpenStack([
    // ...
    'requestOptions' => [
        'stream' => true,
    ],
]);

$objectStore = $streamedOpenstack->objectStoreV1([
    // Give the identity service from the non streamed OpenStack
    //  so it doesn't stream requests for authentication
    'identityService' => $openstack->identityV2(),
]);
```

[...]

PR #299, opened 2020-02-22
Allow user to provide Guzzle options in requests

chapa:
Related to #292

### DIFF
--- a/doc/services/object-store/v1/objects.rst
+++ b/doc/services/object-store/v1/objects.rst
@@ -32,6 +32,9 @@ Download an object
 As you will notice, a Stream_ object is returned by this call. For more information about dealing with streams, please
 consult `Guzzle's docs`_.
 
+By default, the whole body of the object is fetched before the function returns, set the ``'guzzle'`` key of parameter
+``$data`` to ``['stream' => true]`` to get the stream before the end of download.
+
 .. _Stream: https://github.com/guzzle/streams/blob/master/src/Stream.php
 .. _Guzzle's docs: https://guzzle.readthedocs.org/en/5.3/streams.html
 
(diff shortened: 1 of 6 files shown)


chapa:
Arf, just realized it could be better to name the option `requestOptions` instead of `guzzle` (like I suggested in #292) to be consistent with `OpenStack::__construct()`.
What do you think @haphan ?

giulioprovasi:
hey, any eta on this ? have same issue and is kinda blocking the usage of this package for me :/

haphan:
> Arf, just realized it could be better to name the option `requestOptions` instead of `guzzle` (like I suggested in #292) to be consistent with `OpenStack::__construct()`.
> What do you think @haphan ?

Thanks @chapa. I think it makes a lot of sense to re-use the key `requestOptions`. Can you help to update the PR? 

I'm also thinking to have an integration test case to cover this new feature. I can help with integration test if you don't have time.

chapa:
Hey, I'm on vacation right now but I'll take a look when I get back.

If you can do an integration test it would be great !

chapa:
Just pushed the commit to rename `"guzzle'` to `'requestOptions'`.

I'm sorry but I don't have the time to dig into integration tests, I didn't even manage to run them because of this error :
```
$ php ./tests/integration/run.php -s=ObjectStore
INFO: =================================================
INFO: Starting objectstore v1 core integration test(s)
INFO: =================================================
PHP Fatal error:  Uncaught RuntimeException: OpenStack\Integration\Objectstore\v1\CoreTest does not exist in [...]/tests/integration/Runner.php:115
[...]
```
Looks like an autoloading problem, but composer.json seems fine and the autoload.php is required in run.php so I don't understand, I didn't look any further

haphan:
Cool. I will run integration tests on my end.

Issue #312, opened 2020-10-07
Update Guzzle to Version 7


jbeales:
Selfish request, but Guzzle 7 will make it so I can use this in a Laravel 8 app much more easily.

(Working on a PR now).

andrecolza:
I need it too!

jbeales:
I added my fork as a vcs repository to my composer.json, and tagged a 3.0.10 "release" on the Guzzle 7 branch, so now my composer install uses that and I was able to do the Laravel 8 update.

It's kind of hacky, and I'll probably have to delete my vendors directory & composer.lock and do a fresh composer install when the official release comes out, but it allows me to keep working.  Also, when it gets deployed it'll let the Guzzle 7 version of this package get a little production testing, (only for object storage, because that's all I'm doing with it at the moment).

haphan:
@jbeales Thanks for the PR. I will try to review it this weekend 😃

haphan:
I have released `v3.1.0` to address this issue. Thank you all

PR #313, opened 2020-10-08
Update Guzzle dependency to ^7.0

jbeales:
PR for #312.

I don't have full access to an Openstack installation - I'm just a client, not a provider - so if someone could run the integration tests that would be great.

The Identity & Object Store integration tests are passing for me. I'm getting an HTTP 400 from the Networks endpoint when trying to run the Compute test, but I'm wondering if my provider, (OVH), has slightly weird endpoints or something.


jbeales:
A bit more background: The main change in Guzzle 7 that affects this project is that the `template_uri()` function was removed. Luckily the functionality from `tempalte_uri()` was moved into its own [uri-template](https://github.com/guzzle/uri-template) package, so I added that to `composer.json` and added a 'uri_template()` method to `Common\Transport\Utils`.

Other than that I had to update the signature of `HandlerStack::create` to match the updated Guzzle `HandlerStack::create()`, and I found tests weren't running properly because of the name of a directory, so I updated the directory name and all references to it.

haphan:
@jbeales For some reason, GitHub is not showing Travis build results here. Can you help to fix all failed unit tests first?

https://travis-ci.org/github/php-opencloud/openstack/builds/734083798

To trigger unit test locally, `vendor/bin/phpunit`

jbeales:
The unit test now passes, (I needed to cast the response body as a string because Guzzle added a parameter type of 'string' to their parse_json() function), and PSR2 fixes are done.

The PHP 7.1 CI test is still going to fail because Guzzle has a minimum PHP version of 7.2.5, so composer never finishes its install. I guess we should bump the PHP version of this if we want to use Guzzle 7.

haphan:
I'm OK with bumping the PHP version to `php>=7.2.5`

jbeales:
Ok. I found a better way to deal with the failing unit test. I'm going to update the PR.

I'll bump the PHP requirement too.

jbeales:
There's the  latest. The problematic `GuzzleHttp\json_decode()` call was replaced with a `\OpenStack\Common\Transport\Utils::jsonDecode()` call like almost every other time we need to decode JSON from a response.

I also bumped the PHP version.

haphan:
@jbeales  Awesome! Can you help to remove php7.1 from `.travis.yml`? Thanks

jbeales:
Done.

haphan:
@jbeales Thank you for your contribution

jbeales:
Thanks!

PR #317, opened 2020-11-04
Fix testing class namespaces

peter279k:
# Changed log

- Fixing testing class namespaces to be compatible `PSR-4` autoloader.

### DIFF
--- a/tests/unit/TestCase.php
+++ b/tests/unit/TestCase.php
@@ -6,7 +6,6 @@ use function GuzzleHttp\Psr7\stream_for;
 use function GuzzleHttp\Psr7\parse_response;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Psr7\Response;
-use Prophecy\Argument;
 
 abstract class TestCase extends \PHPUnit\Framework\TestCase
 {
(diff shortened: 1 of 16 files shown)


tests/unit/TestCase.php
```diff
@@ -6,7 +6,6 @@
 use function GuzzleHttp\Psr7\parse_response;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Psr7\Response;
-use Prophecy\Argument;
```
peter279k, line 9: It's declared, but not used. It should be removed.

haphan:
@peter279k  thanks for this pr. Does it break anything or more of inconsistency issue?

peter279k:
This will not break anything and it's only for testing when PHPUnit does unit tests and load namespaces autoloading.

haphan:
I mean before this fix, what was not working?

peter279k:
When using the [refactor](https://github.com/rectorphp/rector), it will present some test classes are not found.

And I fix them to be compatible with PSR-4 autoloader.

haphan:
Let me do a quick review later this week

k0ka:
After 3 years of quick review we were able to merge it =).
https://github.com/php-opencloud/openstack/pull/383
Thank you for your PR.

Issue #286, opened 2019-07-16
Deleting a storage object created via createLargeObject() won't delete the segments that make up the file


dsnopek:
If you create an object via `createLargeObject()` and then delete it via this library, it won't actually delete all the segments that make up the object, just the manifest.

I believe there is a way to do the delete request such that OpenStack will delete the segments for you when deleting the object, but I believe that depends on your OpenStack instance having particular middleware. I'm not 100% on either of those points, unfortunately. :-/

In any case, it is possible to ask a manifest where it's segments are, by doing a HEAD on the object and looking at the 'X-Object-Manifest' header. Unfortunately, there isn't a super easy way to get that header - see #285 

Currently, in my application, I have a wrapper around delete that first does a HEAD on the object, checks the 'X-Object-Manifest' header, then lists and deletes all the segments before deleting the manifest. However, it would be really, really great, if this library could handle that automatically! There is builtin support for uploading large objects (via `createLargeObject()`) so IMO there should be support for deleting them as well.

dsnopek:
A little bit of research follows...

Looking at the code for python-swiftclient (which I know does magically delete the segments for a large object) it appears to grab the 'X-Object-Manifest' header and delete the segments itself (as opposed to asking Swift to do it on the server-side):

https://github.com/openstack/python-swiftclient/blob/2fcd4d872713dc30e7352845c37515280f1d21ab/swiftclient/service.py#L2515

For SLO objects (up until now we're talking about DLO objects) it does ask Swift to delete the underlying segments for it, using a special query string parameter (`"multipart-manifest=delete"`). I don't necessarily think this library needs to support deleting SLO objects, since it only supports creating DLO ones, but that could be a nice feature at some point.

Another interesting thing about the Python implementation, is that it uses a thread pool to delete the segments, which would probably be great for performance. If you have 500 segments and each request takes 500ms, it would take 250 seconds to do them serially, but only 50 seconds if done in parallel with 5 threads.

dsnopek:
Based on the above research, I'd like to propose the following:

1. A solution to access extra object headers be implemented in #285 
2. The object `delete()` method get a new argument like `delete(delete_segments=FALSE)` which, if set to true, will grab the 'X-Object-Manifest' header via whatever gets implemented in #285, and then serially deletes the segments before deleting the manifest (doing it in parallel with a thread pool could be done manually by the library user if they choose)

[...]

PR #319, opened 2020-11-04
Left-pad large object segment names with zeros.

### LINKED ISSUE #318: Swift Large Object Segments Sort Incorrectly
If a Large Object uploaded to Swift has more than 10 segments they do not sort correctly, so when they are downloaded the file is re-assembled in the wrong order. [See demo](https://github.com/jbeales/swift-large-object-ordering-demo).

According to the [Swift docs](https://docs.openstack.org/swift/latest/overview_large_objects.html#module-swift.common.middleware.dlo):

> [segments] ...sort in the order in which they should be concatenated. Object names are sorted lexicographically as UTF-8 byte strings.

We are naming the segments with integers, so segments get the names 1, 2, 3.... 9, 10, 11, 12, etc. When Swift re-assembles the large object to be downloaded the segments are ordered 1, 10, 11, 12... 2, 3, 4, etc.

The python-based command-line tools get around this problem by [left-padding the segment name with zeros](https://github.com/openstack/python-swiftclient/blob/master/swiftclient/service.py#L2171) so they sort numerically. They always use a string length of 8 characters.

I'm working on a PR to fix this.

jbeales:
This fixes #318 by using a similar strategy as the official python command-line client. The segment name is left-padded with zeros so when the segment names are sorted on download they end up in the right order.

Instead of always using a length of 8 for the segment name like the command-line client, I checked the string length of the total number of segments, (which will be the name for the last segment), and added one. This way there's no client-imposed limit on the number of segments, at the expense of having to `sprintf()` twice.

### DIFF
--- a/src/ObjectStore/v1/Models/Container.php
+++ b/src/ObjectStore/v1/Models/Container.php
@@ -220,13 +220,15 @@ class Container extends OperatorResource implements Creatable, Deletable, Retrie
             $service->createContainer(['name' => $segmentContainer]);
         }
 
-        $promises      = [];
-        $count         = 0;
-        $totalSegments = $stream->getSize() / $segmentSize;
+        $promises          = [];
+        $count             = 0;
+        $totalSegments     = $stream->getSize() / $segmentSize;
+        $segmentNameLength = strlen( (string) $totalSegments) + 1;
+        $segmentNameFormat = sprintf('%%s/%%0%dd', $segmentNameLength);
 
         while (!$stream->eof() && $count < $totalSegments) {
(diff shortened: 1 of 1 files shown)


jbeales:
I can write a test for this. I would do it by modifying the create_large_object.php sample and updating the ObjectStore/v1/CoreTest's objects method. Do you want that added to this PR, or should it be separate?

jbeales:
I put tests here for now: https://github.com/jbeales/openstack/tree/issue-318-tests The tests currently leave behind a container of segments. If I try to delete the segment container I get a HTTP 409. Maybe because it still has segments in it, (see #286).

PR #323, opened 2020-12-22
Update PHP to 8.0

chindit:
Simple update to supporte PHP 8.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ matrix:
     - php: 7.2
     - php: 7.3
     - php: 7.4
+    - php: 8.0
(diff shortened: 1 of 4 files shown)


ChristophWurst:
```
PHP Fatal error:  Array and string offset access syntax with curly braces is no longer supported in /home/travis/build/php-opencloud/openstack/vendor/jakub-onderka/php-parallel-lint/src/Settings.php on line 145
```

the parallel linter needs an update

k0ka:
The library supports php 8 since version 3.2 - https://github.com/php-opencloud/openstack/pull/325
So I guess we can close the PR

PR #324, opened 2021-01-05
Replace the deprecated linter package

ChristophWurst:
… with the recommended replacement. This will hopefully easy the
adoption of php8.

https://github.com/php-opencloud/openstack/pull/323 failed because of the linter.


ChristophWurst:
It's all done in https://github.com/php-opencloud/openstack/pull/325

PR #331, opened 2021-04-23
Fix Call to undefined function GuzzleHttp\Psr7\modify_request()

alcalyn:
I needed to do this change in order to prevent this error `Call to undefined function GuzzleHttp\Psr7\modify_request()`.

This PR removes deprecated calls to guzzle functions.

I checked all the code, and this was the only missing call to `GuzzleHttp\Psr7\` deprecated functions. I updated the unit tests.


haphan:
Thanks @alcalyn 🚀

PR #342, opened 2022-01-13
Ability to override catalog URL, fix #341

### LINKED ISSUE #341: Support object_storage_url override
Generally, the object_storage_url is built from `OS_AUTH_URL`, `OS_PROJECT_ID` and other similar variable.
Still, the Python Swift client allows for another alternative: [OS_STORAGE_URL](https://github.com/openstack/python-swiftclient/blob/master/swiftclient/client.py#L785)

This allows to use a distinct storage-url unrelated from said variable.

This is useful (= actually _necessary_) in the context of **application credentials,** where an application credential is generated within a given project, but used to access an object container in another project (whose ACLs have been configured). In such a case we can't rely on the URL generation logic.

This is need to give a use for cross-projects https://github.com/php-opencloud/openstack/pull/336
Let's remind that some Openstack hosting solution do not offer multiple users, but only multiple projects, making cross-project application credentials the only secure way to grant access to an object container which implies supporting the OS_STORAGE_URL override.

drzraf:
Sample `$options` that can be passed to `Openstack\Openstack` (in YAML):
```yaml
                catalog_overrides:
                    - name: swift
                      type: object-store
                      region: foobar
                      interface: public
                      url: 'https://foo.bar.cloud:443/v1/AUTH_CUSTOM_ID'
```


haphan:
The python implementation allows overrides to happen at the Swift client level. 

What do you think if we can learn from that pattern to avoid coupling Identity service with the S3 client?

drzraf:
I'm sorry, which coupling are you referring to?

In this PR the endpoint URL override has been made very generic (probably much much more than what need in real-world scenario that would mostly apply to swift and cross-project application credentials or, at best, cross-project ACL).

It's good because, well, it's generic (and, as such, does not touch swift-specifics or tight together different level of components) but may be seen as less obvious to configure than the straight `OS_STORAGE_URL` environment variable that Python swiftclient supports.

drzraf:
Dears, I would be happy to see this in.

k0ka:
Hello, 

Can you add some documentation and tests for this feature?

drzraf:
Sadly not. I'm current not working on this and can't find the time to dig into this again.

drzraf:
@k0ka : Doc:
> Generally, resources and credentials match are expected to match (same project id). But in some cases user credentials may  grant access to a container from another project (different OS_STORAGE_URL) the user originates from.
> In order to bypass the usual automatic URL/resource derivation logic, `catalog_overrides` can be used to specify an alternative `url` that will apply as a replacement for a given service.
> The service to match is specified by either at least its `name` or its `type` (or both) and optional additional restrictions can be added based on the `region` and/or the `interface`.


For example:
```yaml
                catalog_overrides:
                    - name: swift
                      type: object-store
                      region: abc
                      interface: public
                      url:  https://abc.com/v1/AUTH_123456
```

will override the `OS_STORAGE_URL` with `url` when requesting the `public` `object-store` service called `swift` located in `abc`. This is necessary if the user wasn't created within the project `123456`

drzraf:
ping?

k0ka:
This PR doesn't have tests and documentation so I can't accept it. I also don't understand its importance to do it by myself.

PR #343, opened 2022-01-28
Fixes issues with guzzle psr7 compliance

jarragon-slash2:
- Changed guzzle final HandlerStack extension to factory class HandlerStackFactory
- Updated old guzzle psr7 utill functions to work with the new Utill
class functions


k0ka:
Hello, 

Thank you for your PR. I cherry picked renaming `HandlerStack` to `HandlerStackFactory`. 
However moving to guzzle `Utils` might break backward compatibility, so we implemented it another way.

PR #348, opened 2022-07-05
Correctly implement the POST credentials method

svenwiltink:
https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=authenticating-with-an-application-credential-detail,create-credential-detail#credentials
specifies that the credentials endpoint expects a credential object and returns one. This patch ensures the data is properly encoded and decoded.


src/Identity/v3/Models/User.php
```diff
@@ -103,4 +103,9 @@ public function listProjects(): \Generator
     {
         return $this->model(Project::class)->enumerate($this->api->getUserProjects(), ['id' => $this->id]);
     }
+
+    public function listCredentials(): \Generator
```
k0ka, line 107: @return \Generator<mixed, \OpenStack\Identity\v3\Models\Credential>

src/Identity/v3/Service.php
```diff
@@ -338,6 +338,16 @@ public function listCredentials(): \Generator
         return $this->model(Models\Credential::class)->enumerate($this->api->getCredentials());
     }
 
+    /**
+     * Returns a generator which will yield a collection of credential objects. The elements which generators yield can
+     * be accessed using a foreach loop. Often the API will not return the full state of the resource in collections;
+     * you will need to use retrieve() to pull in the full state of the remote resource from the API.
+     */
```
k0ka, line 345: @return \Generator<mixed, \OpenStack\Identity\v3\Models\Credential>

src/Identity/v3/Params.php
```diff
@@ -195,6 +195,15 @@ public function userIdQuery(): array
         ];
     }
 
+    public function userIdQueryUnderscore(): array
```
k0ka, line 198: Why creating new function? 
As far as I see https://docs.openstack.org/api-ref/identity/v3/#credentials 
`POST /v3/credentials` and `GET /v3/credentials` are using the same parameter type `user_id`.

src/Identity/v3/Api.php
```diff
@@ -709,6 +710,15 @@ public function getCredentials(): array
         ];
     }
 
+    public function getUserCredentials(): array
```
k0ka, line 713: There is not need to create a new definition there. We already have `getCredentials()` and can just reuse it by adding the `userId` parameter.

k0ka (changes requested):
Also we will need integration test to check if this parameter really works in OpenStack API.

PR #362, opened 2023-07-20
set alias for image list params, added sort param

matthi4s:
I added alias/`sentAs` definitions to some parameters of the ImagesV2 `getImages()` function which was missing and also added the `sort` parameter. https://docs.openstack.org/api-ref/image/v2/index.html?expanded=list-images-detail#list-images

The `sortKey` and `sortDir` parameters are abstract and also used in the BlockStorageV2 API, but from what I could find they are not really supported there: https://docs.openstack.org/api-ref/block-storage/v2/?expanded=list-snapshots-with-details-detail#list-snapshots-with-details

They are supported in the backups method, but from what I could find, this is not implemented in this library: https://docs.openstack.org/api-ref/block-storage/v2/?expanded=list-backups-with-details-detail#list-backups-with-details
But they use the same alias there and the `sort_key` and `sort_dir` alias follow the naming scheme, so I think it's safe to change it in the abstract class.


k0ka (changes requested):
Surprisingly BlockStorage does use `sortKey` and `sortDir`. 
I added fixes for it and tests.
You might add tests for `sort` and other fields

PR #366, opened 2023-09-22
format segment name, close #365

### LINKED ISSUE #365: The createLargeObject() does not work correctly for segment count 10 and more
In the method `createLargeObject()` there is on this [line](https://github.com/php-opencloud/openstack/blob/aef36abf773bd1992cc416a4c8da951743c1d8e4/src/ObjectStore/v1/Models/Container.php#L229) the following construct

```php
'name'          => sprintf('%s/%d', $segmentPrefix, ++$count),
```

however, when there is more than 9 parts (segments) the resulting large file will not be restored correctly, as the Swift object storage orders the segments by filename...

As stated [here in Swift docs](https://docs.openstack.org/swift/latest/overview_large_objects.html): **_All the object segments need to be in the same container, have a common object name prefix, and sort in the order in which they should be concatenated. Object names are sorted lexicographically as UTF-8 byte strings._**

The current code produces something like this
```
segmentPrefix/1
segmentPrefix/10
segmentPrefix/11
segmentPrefix/2
segmentPrefix/3
```

and should produce

```
segmentPrefix/0001
segmentPrefix/0002
segmentPrefix/0003
:
:
segmentPrefix/0010
segmentPrefix/0011
```

therefore I suggest to change the code to this:

```php
'name'          => sprintf('%s/%05d', $segmentPrefix, ++$count),
```

vpecinka:
For Swift storage and large objects numbers has to be padded with zero for correct segment order.


k0ka:
Thank you for the PR. 

Your current implementation has not obvious limit of 10000 segments. I guess we should either assert it and throw an exception or add another parameter `$data['segmentCountFormat']` to `createLargeObject` with default value `%05d`. 

I also want to add an integration test for this case.

vpecinka:
added the config option `segmentIndexFormat` with default value `%05d` and fixed the unit test.

vpecinka:
also added `isValidSegmentIndexFormat()` to check proper format (has to produce different values for given format and different values, plus result has to have same string length)

PR #367, opened 2023-09-24
Add .gitattributes file

alamirault:
Hello,

This PR add `.gitattributes` with `export-ignore` in order to reduce vendor size

More info here:
https://php.watch/articles/composer-gitattributes#export-ignore

With this change, vendor size goes from 5MB to 1.1MB

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,10 @@
+/.github export-ignore
+/doc export-ignore
+/samples export-ignore
+/tests export-ignore
+/.gitattributes export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,10 @@
+/.github export-ignore
+/doc export-ignore
+/samples export-ignore
```
alamirault, line 3: I'm not sure if we must exclude this folder for DX.

Size is 1.2M

k0ka:
Thank you.

PR #336, opened 2021-08-05
[WIP] Application credentials

smarcet:
this feature branch adds the 

application credentials to identity api

depicted here 

https://docs.openstack.org/keystone/latest/user/application_credentials.html

PR #380, opened 2024-01-07
Application credentials

k0ka:
- add endpoints to create/get/remove application credentials
- add token creation using application credentials

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -7,6 +7,10 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   tests:
     if: |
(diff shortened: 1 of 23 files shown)


drzraf:
Glad this got merged. Still could you consider the comment/commit mentioned at https://github.com/php-opencloud/openstack/pull/336#issuecomment-1004490477
> - application credentials may not request a scope, so the payload sent to /tokens must be modified according to the authentication method
> - Fix an error message regarding the supported authentication methods.

k0ka:
Hello.

1. There are a lot of mutually exclusive parameters in `/v3/auth/tokens` endpoint (https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=authenticating-with-an-application-credential-detail#password-authentication-with-unscoped-authorization), so I guess we should either completely validate parameters or leave it as is. I don't see a reason to only tackle the `scope` parameter. 
2. Yes, I added mention of `application_credential` to the error if no required parameters are found - https://github.com/php-opencloud/openstack/pull/380/files#diff-740662dc3353bdc471612b7e58007cf3be1ec5a34663969542b382dc2a5859b1R108

PR #382, opened 2024-01-07
Handler stack factory

k0ka:
cherry picked from https://github.com/php-opencloud/openstack/pull/343

### DIFF
--- a/src/Common/Service/Builder.php
+++ b/src/Common/Service/Builder.php
@@ -6,10 +6,11 @@ namespace OpenStack\Common\Service;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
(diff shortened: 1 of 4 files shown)


hype09:
Hey there! The removal of the `HandlerStack` class causes a build failure in the project I referenced above - was this an intended BC break in the 3.6 minor version?

k0ka:
Hello,

No, of course it wasn't intended. Sorry for that. I didn't expect anyone would use this internal, undocumented (and quite lame) method. 
Anyway, returned it in https://github.com/php-opencloud/openstack/releases/tag/v3.6.1

hype09:
thank you very much for fixing it this quickly! I'm not sure what this class is being used for in that project, just stumbled on it by accident today when a pipeline failed 😄

PR #404, opened 2024-02-09
Fix the memory limit check

### LINKED ISSUE #403: [Question, potentially a bug] How to debug requests when creating a server
Hi there, I'd appreciate some assistance on the following error that's produced by the SDK:

```
[09-Feb-2024 07:13:20 UTC] HTTP Error
~~~~~~~~~~
The remote server returned a "400 Bad Request" error for the following transaction:
 
Request
~~~~~~~
POST /compute/v2.1/servers HTTP/1.1
User-Agent: GuzzleHttp/7
Content-Type: application/json
Host: something
 
Response
~~~~~~~~
HTTP/1.1 400 Bad Request
Date: Fri, 09 Feb 2024 07:13:20 GMT
Server: Apache/2.4.52 (Ubuntu)
OpenStack-API-Version: compute 2.1
X-OpenStack-Nova-API-Version: 2.1
Vary: OpenStack-API-Version,X-OpenStack-Nova-API-Version
Content-Type: application/json; charset=UTF-8
Content-Length: 137
x-openstack-request-id: something
x-compute-request-id: something else
Connection: close
 
Further information
~~~~~~~~~~~~~~~~~~~
Please ensure that your input values are valid and well-formed. Visit http://docs.php-opencloud.com/en/latest/http-codes for more information about debugging HTTP status codes, or file a support issue on https://github.com/php-opencloud/openstack/issues.
```

This error is produced with code as follows:
```php
// Creates a new user on the OpenStack server and then establishes a new OpenStack connection instance using the new user's credentials.
$OpenStack = $this->authenticateAsClient();

// Now create a new network for the server to be attached to
$network = $OpenStack->networkingV2()->createNetwork([
    'name'         => "Some name",
    'adminStateUp' => true,
]);

// Finally create the server
$server = $OpenStack->computeV2()->createServer([
    'name'     => "Some name",
    'imageId'  => "The image ID",
    'flavorId' => "The flavor ID",
    'networks'  => [
        ['uuid' => $network->id]
    ],
]);
```

I am fairly new to OpenStack and I am working on a project with it, so I apologize if there's something I am blaitantly doing wrong here, but the logged message isn't particularly clear on *what* the problem is & searching "debug" on the docs returns no results.

Thanks!

BelleNottelling:
Closes #403

When you query `ini_get('memory_limit')` you don't always get an integer response, it returns exactly what is written inside of the PHP `.ini` file file which typically includes a shorthand value. (`K`, `M`, or `G`, all case-insensitive).

My server has the default PHP limit of `128M` which was incorrectly being evaluated as not enough by the check.

This PR corrects the behavior by first evaluating the set memory limit, including the shorthand byte values and then returns it as an integer which can then be properly compared against the message's body size.

### DIFF
--- a/src/Common/Error/Builder.php
+++ b/src/Common/Error/Builder.php
@@ -64,7 +64,8 @@ class Builder
     }
 
     /**
-     * Reads the PHP memory limit from the ini files and then converts it into an integer, handling PHP's shorthand byte values
+     * Reads the PHP memory limit from the ini file and then converts it into an integer, handling PHP's shorthand byte values.
+     * 
      * @see https://www.php.net/manual/en/faq.using.php#faq.using.shorthandbytes
      * 
      * @return int
--- a/src/Common/Error/Builder.php
+++ b/src/Common/Error/Builder.php
@@ -54,7 +54,7 @@ class Builder
      */
     private function linkIsValid(string $link): bool
     {
-        $link = $this->docDomain.$link;
+        $link = $this->docDomain . $link;
 
         try {
             return $this->client->request('HEAD', $link)->getStatusCode() < 400;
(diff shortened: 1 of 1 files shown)


k0ka:
Well, I don't think it even make sense to check memory_limit there. It looks like some old temporary fix. I guess it's better to show body only for jsons and truncate it by like 5kb.

BelleNottelling:
> Well, I don't think it even make sense to check memory_limit there. It looks like some old temporary fix. I guess it's better to show body only for jsons and truncate it by like 5kb.

I don't think it makes sense either, but when it comes to someone else's codebase (and one that I am unfamiliar with) I tend to lean towards fixing a bug rather than modifying existing behavior. I honestly am unable to think of any valid situation for this check to exist at all truthfully

PR #406, opened 2024-03-27
HTTP header names are case-insensitive

drzraf:
I guess a Guzzle update or the swift provider HTTP server or a Swift update (or an intermediary proxy) changed the case of the HTTP headers and... `getMetadata()` started being empty what drove me crazy until I understood it.

### DIFF
--- a/src/ObjectStore/v1/Models/MetadataTrait.php
+++ b/src/ObjectStore/v1/Models/MetadataTrait.php
@@ -13,7 +13,7 @@ trait MetadataTrait
         $metadata = [];
 
         foreach ($response->getHeaders() as $header => $value) {
-            if (0 === strpos($header, static::METADATA_PREFIX)) {
+            if (0 === stripos($header, static::METADATA_PREFIX)) {
                 $name            = substr($header, strlen(static::METADATA_PREFIX));
                 $metadata[$name] = $response->getHeader($header)[0];
             }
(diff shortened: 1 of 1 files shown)


src/ObjectStore/v1/Models/MetadataTrait.php
```diff
@@ -13,9 +13,12 @@ public function parseMetadata(ResponseInterface $response): array
         $metadata = [];
 
         foreach ($response->getHeaders() as $header => $value) {
-            if (0 === strpos($header, static::METADATA_PREFIX)) {
+            if (0 === stripos($header, static::METADATA_PREFIX)) {
                 $name            = substr($header, strlen(static::METADATA_PREFIX));
                 $metadata[$name] = $response->getHeader($header)[0];
+                if ($name !== strtolower($name)) {
```
k0ka, line 19: This if is not needed. And it makes the code slower as `strtolower` executed twice.

k0ka (changes requested):
need a unit test for this case

src/ObjectStore/v1/Models/MetadataTrait.php
```diff
@@ -13,9 +13,12 @@ public function parseMetadata(ResponseInterface $response): array
...
+                    $metadata[strtolower($name)] = $response->getHeader($header)[0];
```
k0ka, line 20: I'm also not sure we need this at all. It might break the backward compatibility as swift metadata would have double keys.

drzraf: It allows my `$container->getMetadata()["temp-url-key"]` code to work indistinctly on both OVH swift 2.33 (and its all-lowercase headers) but also on another provider's Swift 2.25 with mixed-case headers' names.

drzraf:
> need a unit test for this case

I'm sorry, I can't enter into such a thing.

PR #422, opened 2025-03-11
Add qos_network_id option to ports create/update 

tduncan8:
add qos_network_id to ports create/update to allow this field to be readable/writeable https://docs.openstack.org/api-ref/network/v2/index.html


k0ka:
Hello, 

Thank you for the PR. Could you create some tests for this feature? At least add a unit test with some real response from the server to make sure it is working properly. Like this:
https://github.com/php-opencloud/openstack/blob/master/tests/unit/Networking/v2/Models/PortTest.php#L50
https://github.com/php-opencloud/openstack/blob/master/tests/unit/Networking/v2/Fixtures/port_get.resp

PR #424, opened 2025-04-30
nimbusoft/flysystem-openstack-swift can't be installed due to use of json-schema 5

evs-xsarus:
Using composer 2.8.8 , it's not possible to require nimbusoft/flysystem-openstack-swift . That package uses openstack which is limited to justinrainbow/json-schema ^5.2. But composer 2.8.8 requires at least ^6.3  . This MR adds support for installing ^6.0 with openstack.


k0ka:
Thank you for contribution. Release with https://github.com/php-opencloud/openstack/releases/tag/v3.13.0

PR #428, opened 2025-05-30
fix integration tests, add epoxy release

k0ka:


k0ka:
It turns out, new version devstack doesn't create cinderv3/volumev3 service - https://review.opendev.org/c/openstack/tempest/+/930296
I'll add fix to tests and docs. We can't change default `blockStorageV3` as it would break backward compatibility. The latest docs are still asking to create `cinderv3`/`volumev3` service - https://docs.openstack.org/cinder/2025.1/install/cinder-controller-install-ubuntu.html

